PR #36, opened 2014-10-20
Translated Workflows Summary & Workflows selects

playmono:


joebordes:
Processed this PR and added the different parts in a series of commits (most are marked with the PR number).

Sorry for taking so long but it was a complex PR :-)

Thanks Adrian

PR #49, opened 2015-03-24
fix in save calendar from workflow event

evolutivo:


omarllorens:
Thank you Lorida, works perfect.

PR #51, opened 2015-04-10
Some translation

playmono:
Coherence in translations and translation of "new condition"


joebordes:
Thank you!!

Have a nice weekend

PR #52, opened 2015-04-15
fix querygen

evolutivo:
fix query generator bug on assigned to of related modules

### DIFF
--- a/include/ListView/ListViewController.php
+++ b/include/ListView/ListViewController.php
@@ -236,7 +236,8 @@ class ListViewController {
                                 $modrel=getTabModuleName($field->getTabId());
                                 $rawValue = $this->db->query_result($result, $i, "smowner".strtolower($modrel));
                                 }
-				else $rawValue = $this->db->query_result($result, $i, $field->getColumnName());				if($module == 'Calendar') {
+				else $rawValue = $this->db->query_result($result, $i, $field->getColumnName());				
+                                if($module == 'Calendar') {
 					$activityType = $this->db->query_result($result, $i, 'activitytype');
 				}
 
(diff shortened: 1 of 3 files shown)


include/ListView/ListViewController.php
```diff
@@ -208,8 +232,11 @@ function getListViewEntries($focus, $module,$result,$navigationInfo,$skipActions
 					if (is_null($field)) continue;
 				}
 				$uitype = $field->getUIType();
-				$rawValue = $this->db->query_result($result, $i, $field->getColumnName());
-				if($module == 'Calendar') {
```
joebordes: I think you have made a mistake here. You have to leave the

```
if($module=='Calendar')
```

and put it below with the $activityType assignment

include/ListView/ListViewController.php
```diff
@@ -440,6 +467,9 @@ function getListViewEntries($focus, $module,$result,$navigationInfo,$skipActions
 						$value = '--';
 					}
 				} elseif($field->getFieldDataType() == 'owner') {
+                                        if($fieldName!='assigned_user_id' && strstr($fieldName,".assigned_user_id"))
```
joebordes: Since you have to modify the file to fix the above comment I would appreciate you added the block delimeters to this "if":

```
if($fieldName!='assigned_user_id' && strstr($fieldName,".assigned_user_id")) {
$value = textlen...
} else {
....
}
```

Thanks

include/ListView/ListViewController.php
evolutivo: I committed the changes. I didn't cancel the row, it was moved at the end of the line that's why you didn't see it, my mistake

joebordes:
Thank you :-)

PR #53, opened 2015-04-17
fix report rel generation

evolutivo:
Fix report generation when choosing uitype 10 field columns(trello bug).


joebordes:
Thanks!

PR #54, opened 2015-04-21
query generator same names different modules problem

evolutivo:


omarllorens:
Good job! Thanks!

evolutivo:
Thank you :)

PR #59, opened 2015-10-08
fix(Header/Login) dynamic way to get favicon

evolutivo:


Smarty/templates/LoginHeader.tpl
```diff
@@ -14,7 +14,7 @@
 <meta http-equiv="Content-Type" content="text/html; charset={$LBL_CHARSET}">
 <meta http-equiv="X-FRAME-OPTIONS" content="DENY" />
 <title>{$APP.LBL_BROWSER_TITLE}</title>
-<link REL="SHORTCUT ICON" HREF="themes/images/favicon.ico">
+<link REL="SHORTCUT ICON" HREF="{php} $logo_login=fetch_logo(3); if ($logo_login!='test/logo/') echo $logo_login; else echo 'themes/images/favicon.ico' {/php}">
```
joebordes: I think this is incorrect, it should be just ```{$FAVICON}``` because the logic of the "if" is already done in the header.php change you do below. Besides you shouldn't be using the PHP directive in smarty it is deprecated and not supported anymore.

Please fix this and I will accept as this is an important change to reduce logs size.

Thanks

evolutivo: I changed the code. Couldn't use header.php for login variables. Used modules/Users/Login.php

joebordes:
Perfect!
Thank you!

PR #60, opened 2015-11-06
update(Assets,Calendar) Dutch translation

Luke1982:
It also included my gitignore in this commit, but I can't delete this.


Luke1982:
dfgdfgdg

.gitignore
```diff
@@ -7,4 +7,6 @@
 /backup
 /user_privileges
 /config-dev.inc.php
+/config.inc.php
```
joebordes: I don't want to add this to coreBOS because we version config.inc.php on production sites. We fork the base project for each of our installs and version the config.inc.php file with the installs preferences.
We then override those preferences with the config-dev.inc.php file for development.

Smarty/templates/modules/Calendar4You/CalendarView.tpl
```diff
@@ -456,6 +456,23 @@ function hideITSEventInfo(){
     jQuery('#event_info_content').html('');
 
 }
+
```
joebordes: I like this idea but not exactly how it is implemented, I don't like the idea of having the toolbar hovering over the calendar. I would prefer for it to disappear and have the calendar adapt to the available space. Based on this code, I gave it a try and almost got it working but it breaks on changes from month/week/day.
I'll have another go at it when I can.

modules/Assets/language/nl_nl.lang.php
```diff
@@ -1,5 +1,4 @@
 <?php
```
joebordes: Integrated

modules/Calendar4You/CalendarUtils.php
```diff
@@ -441,12 +441,14 @@ function transferForAddIntoTitle($type, $row, $CD) {
...
-
+	if ($CD["fieldname"] == "location")  {
+		$value = "<a href='https://www.google.nl/maps/place/".$row['location']."' target='_blank'>".$row['location']."</a>";
+	}
 	if ($type == "1")
 		return $Cal_Data[1];
 	else
-//		return "<br><b>".$Cal_Data[0]."</b>: ".$value;
-		return '<table><tr><th>'.$Cal_Data[0].':</th><td onmouseover="vtlib_listview.trigger(\'cell.onmouseover\', $(this))" onmouseout="vtlib_listview.trigger(\'cell.onmouseout\', $(this))">'.$value.'</td></tr></table>';
+		return '<br><b>'.$Cal_Data[0].'</b>: <span onmouseover="vtlib_listview.trigger(\'cell.onmouseover\', $(this))" onmouseout="vtlib_listview.trigger(\'cell.onmouseout\', $(this))">'.$value.'</span>';
```
joebordes: This breaks the tooltip feature. We changed from the direct text to the table when we added support for tooltips in the calendar event. You can define a tooltip for accounts (for example) and when you hover your mouse over the account name in the event it will show you the details of the account. Tooltip needs the table to work correctly, so to be able to apply this change I would need you to make an individual PR with some changes to tooltip to work correctly.

modules/Calendar4You/CalendarUtils.php
```diff
@@ -441,12 +441,14 @@ function transferForAddIntoTitle($type, $row, $CD) {
 		$value = getTranslatedString($Cal_Data[1],'Calendar');
 	else
 		$value = $Cal_Data[1];
-
+	if ($CD["fieldname"] == "location")  {
```
joebordes: To accept this I would need to have the google URL configurable. Maybe you can create an individual PR with a Global Variable?

modules/Calendar4You/language/nl_nl.lang.php
```diff
@@ -7,72 +7,74 @@
  * All Rights Reserved.
```
joebordes: integrated

modules/CobroPago/language/nl_nl.lang.php
```diff
@@ -99,8 +99,7 @@
 	'Activities' => 'Activiteiten',
```
joebordes: integrated

modules/VtigerBackup/VtigerBackup.php
```diff
@@ -26,7 +26,12 @@
  * @author MAK
  */
 class VtigerBackup {
-	private $folderList = array('storage','test','user_privileges');
+	// 'storage' folder removed from backup, the CBX one is too large and will run into memory limits
```
joebordes: I can't accept this as copying storage files is fundamental. You now have an external backup method which will do big storage directories correctly.

joebordes:
Please create individual PR for each logical concept. Mixing them all together makes it really hard to process. To do that you create individual branches for each "thing" you need to work on.

Thank you very much for sharing :-)

Smarty/templates/modules/Calendar4You/CalendarView.tpl
Luke1982: Yes that was my first thought as well. Unfortunately the 'fullcalendar' plugin needs to refire to calculate widths in this case (widths are not relative but calculated absolutely). Another solution would maybe be to wrap the entire content, make it 120% of the screen size and shift it left and right via jQuery.

modules/Calendar4You/CalendarUtils.php
Luke1982: I've got the tooltip working in this code, but it is a little 'jumpy' as in nervous to appear/disappear. The problem with the tabled layout is when you have a lot of appointments next to each other (like in my use case). Any text becomes unreadable in this case.

modules/Calendar4You/CalendarUtils.php
Luke1982: Ah this would be a nice chance for me to play with the Global Variable. Will test this and report back.

modules/VtigerBackup/VtigerBackup.php
Luke1982: Yes this was only meant for my personal fork, not as a suggestion for coreBOS. I'm not familiar with the external backup though.. Will research

Luke1982:
I know, this was not my intention, I just don't know how to work with GIT that well.. Will update my knowledge and work on this.

joebordes:
I know! The initial learning curve is a little steep, but it is REALLY worth it, It is an incredible tool.
Keep asking, I'll help all I can

modules/VtigerBackup/VtigerBackup.php
joebordes: https://github.com/tsolucio/corebos/tree/master/cron/modules/VtigerBackup
https://github.com/tsolucio/corebos/blob/master/build/changeSets/cronbackup.php

modules/Calendar4You/CalendarUtils.php
joebordes: Let's move this discussion to the forum. On the thread you opened about it. I understand and have seen the problem, but I couldn't dedicate time to looking into it. I was thinking of making the layers "elevate" when you click on it or something like that. I also wanted to look into upgrading full calendar, we are using an old version and maybe a newer version does this better....

Smarty/templates/modules/Calendar4You/CalendarView.tpl
joebordes: Let's keep investigating and learning. Most of the calendar functionality is new to me too.

PR #61, opened 2015-11-19
fix(popup function in vtlibutils and calendar4you adding other date u…

evolutivo:
…itypes to show in calendar)

### DIFF
--- a/include/utils/VtlibUtils.php
+++ b/include/utils/VtlibUtils.php
@@ -647,8 +647,10 @@ function vtlib_purifyForSql($string, $skipEmpty=true) {
 }
 
 function getvtlib_open_popup_window_function($popupmodule,$fldname,$basemodule) {
+    if(file_exists('modules/'.$popupmodule.'/'.$popupmodule.'.php')){
 	include_once 'modules/'.$popupmodule.'/'.$popupmodule.'.php';
 	$mod = new $popupmodule();
+    }
 	if (method_exists($mod, 'getvtlib_open_popup_window_function')) {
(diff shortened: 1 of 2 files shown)


include/utils/VtlibUtils.php
```diff
@@ -647,8 +647,10 @@ function vtlib_purifyForSql($string, $skipEmpty=true) {
 }
 
 function getvtlib_open_popup_window_function($popupmodule,$fldname,$basemodule) {
+    if(file_exists('modules/'.$popupmodule.'/'.$popupmodule.'.php')){
 	include_once 'modules/'.$popupmodule.'/'.$popupmodule.'.php';
 	$mod = new $popupmodule();
+    }
 	if (method_exists($mod, 'getvtlib_open_popup_window_function')) {
```
joebordes: Shouldn't this "if" be inside the "file_exists if"?  If you don't load the file $mod will be undefined and the method_exists will emit a warning at least.
Please, move the "method_exists if" inside the "file_exists if"

joebordes:
Please separate this into two commits/PRs. Both file changes are totally unrelated.

joebordes:
I deliberately left out uitype 6 and 23. The SQL is totally intentional.
6 and 23 really shouldn't be used and they are not supported uitypes. They are hard coded for calendar. I would prefer to create a new uitype, but the recommended way to do time fields is with uitype 14.
Why do you need these uitypes there?

evolutivo:
We had some custom modules like Task module which used uitype 6 and 23 dates.  Maybe it's better for us to change the date fields of those modules and leave the queries as they are. Thanks

PR #66, opened 2015-12-23
feat(Contacts) Contact Hierarchy

casati-dolibarr:
DetailView.tpl & DetailViewUI.tpl are my original 5.4 changes (5.4 file plus changes before CoreBOS).

So most of the changes YOU did after vt54 to cb54 release are NOT included there! This needs most of your professional scrutinity.

All others should be easy to understand.

Code is mostly bracketed by '// pag' pairs


joebordes:
I analyzed the PR and extracted the Contact Hierarchy functionality.
You will find it in commit:

https://github.com/tsolucio/corebos/commit/91b1a6c7b75b38c8a369308463f325667e434d97

The rest of the changes won't be incorporated. I like the idea of hiding and compacting the comments section, I may look into that one in the future but I need a generic way of doing the tabs not hard code it for one module.

Anyway, thank you very much :-)

Happy New Year

PR #70, opened 2016-01-19
fix(PackageImport.php) attach module to Tools when parenttab doesn't …

kduqi:
…exist


joebordes:
Looks good. Thank you!!

PR #71, opened 2016-01-25
feat(ButtonList) adds classes and id's to the buttonlist template

Luke1982:
This way, theme designers can easily target these individually (also
better javascript targeting can be achieved).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /user_privileges
 /config-dev.inc.php
 /modules/Calendar4You/googlekeys
+/config.inc.php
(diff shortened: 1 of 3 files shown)


joebordes:
I cherry-picked the first two commits and left out the config.inc.php changes because you can use config-dev.inc.php for your changes and config.inc.php for your development installs.

You now appear as an official contributor of coreBOS :-)
Thank you very much!!

Luke1982:
Ah, I was trying to cherry-pick during the creation of the pulls but now luck. Thanks.

PR #74, opened 2016-02-09
Search for config-dev.inc.php without silencing errors. 

kikojover:
In some php installations when using @include_once, application fails with an apache segmentation fault.

Change searching config-dev.inc.php with file_exists


joebordes:
I am looking at the commit on the cell phone but it seems like there is an unbalanced quote in the file_exists. Could be?

PR #76, opened 2016-05-12
Customblocks

kikojover:
Add feature to customize module Blocks.
If *Smarty/templates/modules/MODULE/BLOCK_LABEL_detail.tpl* is present, overrides default detail block view, same for edit with *Smarty/templates/modules/MODULE/BLOCK_LABEL_edit.tpl*

MODULE is module name.
BLOCK_LABEL is the label of the block you want customize.

Note that is the developer job to recreate full structure for block, for that you can use $FIELDS array with crmobject->custom_fields in it, in the Smarty template.

Modules without DetailView.php script are'nt adapted, also:
- Email
- Users
- Calendar
- Webmails

Because it's DetailView.php script differs so much than default one.

Enjoy !

### DIFF
--- a/Smarty/templates/DetailView.tpl
+++ b/Smarty/templates/DetailView.tpl
@@ -329,6 +329,7 @@ function sendfile_email()
 																						</tr>
 																					{/if}
 																				</table>
+                                                                                                                                                                
 																				{if $header neq 'Comments'}
 																					{if $BLOCKINITIALSTATUS[$header] eq 1}
 																						<div style="width:auto;display:block;" id="tbl{$header|replace:' ':''}" >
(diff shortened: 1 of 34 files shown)


joebordes:
It looks very good Kiko. I am going to accept the PR, test it and put up some test code in the wiki.
There are two things I see missing:

1.- There are some modules that use the Smarty/templates/CreateView.tpl, for example, Documents:

https://github.com/tsolucio/corebos/blob/master/modules/Documents/EditView.php#L245

and many custom modules, so you will have to add the same changes you put in salesEditView.tpl to this file

2.- The DetailView.php changes need to be applied also to the new module base files:

vtlib/ModuleDir/DetailView.php

You can add those two changes directly into coreBOS. Commit them directly.

Thank you very much for sharing this great idea :-)

joebordes:
http://corebos.org/documentation/doku.php?id=en:devel:add_editdetail_block

kikojover:
Done !! Also i've modified custom EditView.php script on Documents.

Thanks for the mention on Wiki :-)

joebordes:
FYI:
In the latest version of the code the custom blocks are loaded even if they have no fields inside:

https://github.com/tsolucio/corebos/commit/78ccb6be8efcef6adf986229de92f4bfff26a940

PR #78, opened 2016-06-20
Albertxhani branch

evolutivo:


modules/Vtiger/duplicate.php
```diff
@@ -0,0 +1,125 @@
...
+if(isset($_REQUEST['module_name']) && isset($_REQUEST['record_id']))
+{
+	$currentModule = vtlib_purify($_REQUEST['module_name']);
+	$record_id = vtlib_purify($_REQUEST['record_id']);
+	
+	require_once 'modules/'.$currentModule.'/'.$currentModule.'.php';
+	$focus = new $currentModule();
+	$focus->retrieve_entity_info($record_id, $currentModule);
+
+	//Duplicate Records that this Record is dependent of
```
joebordes: We don't need this block. Our specification is to duplicate dependent and related modules, not directly related records. Please comment this piece of code as it may be useful in the future.

modules/Vtiger/duplicate.php
```diff
@@ -0,0 +1,125 @@
...
+			require_once "modules/" . $module ."/". $module .".php";
+			$entity = new $module();
+			$entity->retrieve_entity_info($value,$module);
+			$entity->save($module);
+			$new_entity_id = $entity->id;
+			$focus->column_fields[$fieldname] = $new_entity_id;
+		}
+	}
+
+	$focus->save($currentModule);
```
joebordes: I think this won't work correctly. If I am correct, when you get information from the database it comes in the database format but when you save it, it has to be in the current user's format, so some fields may be saved incorrectly. You can try this with special characters, date and numeric fields.
http://corebos.org/documentation/doku.php?id=en:devel:saverecordfromwithin

If I am correct you have to pass column_fields through DataTransform::sanitizeRetrieveEntityInfo like this:

https://github.com/tsolucio/corebos/blob/master/modules/com_vtiger_workflow/tasks/VTUpdateFieldsTask.inc#L112

joebordes:
@Albert, I have enhanced the steps to create a Business Mapping on the wiki:

http://corebos.org/documentation/doku.php?id=en:adminmanual:businessmappings#how_a_programmer_can_add_a_business_mapping_type

- I would like you to create the business mapping following these steps. Call it DuplicateRelations and search for {ModuleName}_DuplicateRelations when looking for the mapping, where {ModuleName} is the name of the module whose record we are duplicating.

- You shouldn't need to load the mapping files, they are system files loaded at the beginning of the launch,

- I would like you to document the XML structure in the Mapping file like the others do.

- I would like you to add another directive in the XML called DuplicateDirectRelations if it is set to true you duplicate the "1" relations like if it is set to false you use the existing relations. The idea is that, normally we will want to duplicate a record and all of it's related records, but the main record we are duplicating will stay related to the same records it currently is. not new ones. It is this loop: 

https://github.com/tsolucio/corebos/pull/78/files#diff-ff2539faf98e599bd0ef754abb142f48R18

what I am trying to say is that this loop will be conditional, if DuplicateDirectRelations is true you will enter this loop, if not you won't and the new record will be directly related to the same parent records.

joebordes:
Please add the license to the new Duplicate.php file.

https://github.com/tsolucio/corebos/blob/master/build/code_license.tpl#L50

evolutivo:
I think Its very profitable. It reduces a lot of repeated work done by programmers

PR #79, opened 2016-07-08
fix(RecycleBin) so when you click empty module whithout selecting any…

kduqi:
… record it just asks for confirmation and empties the module


joebordes:
Thanks!

PR #83, opened 2016-08-08
feat(Business Map) Add Module Set Mapping

evolutivo:


include/Webservices/ConvertLead.php
```diff
@@ -264,4 +266,73 @@ function vtws_updateConvertLeadStatus($entityIds, $leadId, $user) {
...
+		$contactid = vtws_getIdComponents($entityIds['Contacts'])[1];
+
+	$bmapname = 'LeadConversion';
+	$cbMapid = GlobalVariable::getVariable('BusinessMapping_'.$bmapname, cbMap::getMapIdByName($bmapname));
+	if ($cbMapid) {
+		$cbMap = cbMap::getMapByID($cbMapid);
+		$modules = $cbMap->ModuleSetMapping()->getFullModuleSet();
+
+		foreach ($modules as $module) {
+			if($module != "Potentials") {
```
joebordes: Shouldn't we avoid here also Accounts and Contacts??

joebordes:
Looks wonderful. I'll give it a try and document it on the wiki soon and get back to you if I run into any issues.

Thanks!!

include/Webservices/ConvertLead.php
evolutivo: This will cause problems only if the user creates Account2Contacts or Contact2Accounts maps. Either way we should prevent that (TBH I didn't think of that)

joebordes: Done 2b887856cfe087cebcfc72fbff68562edd72c6d5
Thanks!

PR #84, opened 2016-08-08
Remove fixed widths on edit, save and delete buttons.

Luke1982:


Luke1982:
I don't understand this at all. I created a fresh branch and commited from that branch. Now this PR is showing commits from january. I will try again, this is not working at all.

PR #87, opened 2016-08-11
Create invoice from so workflow

Luke1982:
This now also works with AJAX action. You'll probably want me to sanitize the date fields, but please guide me a bit on how to do so.


joebordes:
I merged your code and took it as a starting point to eliminate one of the many outstanding things I have on my endless to-do list.

I constructed a generic inventory module converter workflow instead of only one specific one.

Let me try to point out some of the important parts for you:

- the workflow applies only for inventory modules:
https://github.com/tsolucio/corebos/blob/master/build/changeSets/add_workflow_convert_inventorymodule.php#L25
'include' => getInventoryModules()

- from_wf
https://github.com/tsolucio/corebos/blob/master/modules/com_vtiger_workflow/tasks/convert_inventorymodule.inc#L35
tells the system we are creating from inside a workflow and support displaytype=4 fields

- lines 36, 37 and 85 take care of the current_user (although I'm not quite sure how that happens)

- lines 75-78 and 82-83 apply the action change just for this operation

- to give users control over the field mappings I use the default system mapping function and the new Business Rules

- finally I added the translation strings and a helper note.

Thank you VERY much. Really, I appreciate this a lot!

Let me know if you have any questions and also how it goes when you try it.

:+1: :

Luke1982:
Thanks, looks like we made some good functionality together. I'll study your code (especially the functions I'm not familiar with) and get back to you if I have questions. I'm in the middle of some projects right now, but when I get the time I'll plan some coreBOS development time and let you know.

PR #91, opened 2016-08-22
Dutch language

Luke1982:
Somehow, I don't understand why, my 'workflow for creating an invoice from SO' commits got into this branch. Didn't even know commits could be in multiple branches. Hope you can cherry pick the commits that are actually language....


joebordes:
Thanks!!

PR #92, opened 2016-08-30
feat(Documents) mass document download action

kduqi:


Smarty/templates/DocumentsListViewEntries.tpl
```diff
@@ -139,6 +139,9 @@
 															<input type="button" name="show" value="{$MOD.LBL_VIEW_EMPTY_FOLDERS}" class="crmbutton small cancel" onClick="fnvshobj(this,'emptyfolder');" title="{$MOD.LBL_VIEW_EMPTY_FOLDERS}">
 														</td>
 													{/if}
+													<td>
```
joebordes: Please convert this into a ListViewBasic link, somthing like this:
https://github.com/tsolucio/corebos/blob/master/build/changeSets/cbupdater_importxml.php#L26

modules/Documents/SaveFile.php
```diff
@@ -58,4 +58,36 @@
 			echo 'file_available';
 		}
 }
+
+if(isset($_REQUEST['act']) && $_REQUEST['act'] == 'massDldCnt')
+{
+	$all_files = $_REQUEST['file_id'];
```
joebordes: add vtlib_purify here

modules/Documents/SaveFile.php
```diff
@@ -58,4 +58,36 @@
...
+	$zipfilename = "cache/Documents.zip";
```
joebordes: add the userid here

cache/Documents{userid}.zip

that way if two or more users are downloading they don't overwrite each other.

modules/Documents/SaveFile.php
```diff
@@ -58,4 +58,36 @@
...
+						$name = @$adb->query_result($result, 0, 'name');
+						$filepath = @$adb->query_result($result, 0, 'path');
+						$name = html_entity_decode($name, ENT_QUOTES, $default_charset);
+						$saved_filename = $pname."_".$name;
+						$zip->addFile($filepath.$saved_filename, $folderName.'/'.$saved_filename);
+					}
+				}
+			}
+		}
+	}$zip->save();
```
joebordes: put the "save" on it's own line

modules/Documents/SaveFile.php
```diff
@@ -58,4 +58,36 @@
...
+	$zip->forceDownload($zipfilename);
+	unlink($zipfilename);
```
joebordes: I would eliminate this just in case the code deletes the file before the download finishes, you have it controlled on line 67 anyway and it is a cache directory which can be cleaned whenever the system administrator wants.

PR #93, opened 2016-08-31
fix(Documents) mass download

kduqi:


joebordes:
Works very nice. I like the way you respect the folder structure inside the .zip.

Thanks!!

PR #96, opened 2016-09-09
Actioncolumn: improve the actioncolumn and prepare it for better theming

Luke1982:
This branch has a few small commits that improve the actioncolumn. There is a fallback for the softed theme (which is also an improvement) and this will make theming easier in the future.


joebordes:
Thanks!

PR #99, opened 2016-09-13
feat(cbMap) Import Type

AlbanaCelepija:


modules/cbMap/processmap/importCSV.php
```diff
@@ -0,0 +1,175 @@
...
+
+//ini_set('display_errors', 'on');
+require_once("config.inc.php");
+include_once('data/CRMEntity.php');
+include_once('modules/cbMap/cbMap.php');
+require_once('include/utils/utils.php');
+require_once('include/database/PearDatabase.php');
+
+global $adb, $log, $root_directory, $current_user;
+$current_user = new Users();
```
joebordes: Please change these two lines to:

```
$current_user = Users::getActiveAdminUser();
```

just in case we run into an install where the admin user 1 has been deleted

modules/cbMap/processmap/importCSV.php
```diff
@@ -0,0 +1,175 @@
...
+    }
+$mapfocus = CRMEntity::getInstance("cbMap");
+$mapfocus->retrieve_entity_info($mapid, "cbMap");
+$mapinfo = $mapfocus->readImportType();
+$updateFld = $mapinfo['target'];
+$matchFld = $mapinfo['match'];
+$options = $mapinfo['options'];
+
+$module = $mapfocus->getMapTargetModule();
+include_once("modules/$module/$module.php");
```
joebordes: this line is not needed, CRMEntity::getInstance below checks the file exists and includes it for you

modules/cbMap/processmap/importCSV.php
```diff
@@ -0,0 +1,175 @@
...
+                    }
+                    elseif (!empty($data[$value]))
+                        $focus->column_fields[$upkey] = $data[$value];
+                    else
+                        $focus->column_fields[$upkey] = $predefined;
+                }
+                $focus->mode = 'edit';
+                $focus->id = $index_result;
+
+                $focus->saveentity($module);  $k1++;
```
joebordes: don't you have to sanitize data here? are special characters or non-iso dates and currency working correctly?

modules/cbMap/processmap/importCSV.php
```diff
@@ -0,0 +1,175 @@
...
+                        }
+                    }
+                    elseif (!empty($data[$value]))
+                        $focus->column_fields[$upkey] = $data[$value];
+                    else
+                        $focus->column_fields[$upkey] = $predefined;
+                }
+                $focus->mode = 'edit';
+                $focus->id = $index_result;
+
```
joebordes: don't you have to sanitize data here? are special characters or non-iso dates and currency working correctly?

modules/cbMap/processmap/importCSV.php
```diff
@@ -0,0 +1,175 @@
...
+                                $focus1->column_fields[$upkey] =$adb->query_result($index_rel,0);
+                                             
+                        }
+                        }
+                        elseif (!empty($data[$value]))
+                        $focus1->column_fields[$upkey] = $data[$value];
+                        else
+                        $focus1->column_fields[$upkey] = $predefined;
+                }
+                $focus1->column_fields["assigned_user_id"]=1;
```
joebordes: this should be:

```
$focus1->column_fields["assigned_user_id"]=$current_user->id;
```

modules/cbMap/processmap/importCSV.php
```diff
@@ -0,0 +1,175 @@
+<?php
+
```
joebordes: Please add the license here: https://github.com/tsolucio/corebos/blob/master/build/code_license.tpl#L50

joebordes:
I would reduce the import script considerably. The way I imagine Business Rules (cbMap) is that each class/map is self contained, with all the functionality inside the class.
So I would see that Import() would have two more methods:

```
initializeImport()
```

this method would do all the first part you have in importCSV, the part about deleting, creating and filling tables

then another method called

```
doImport()
```

would do the second part of the importCSV script. the part where it runs through the work to be done and updates or creates

then I would probably even take it a step further and do all this in the processMap method:

```
function processMap($arguments) {
	$this->convertMap2Array();
	if xml not ok return error
	$this->initializeImport();
	$this->doImport();
	return 'ok';
}
```

this would leave the importCSV script with only a few lines to load the map you want and execute it, making it not necessary, just something that you can put up in the documentation page as a script to be created by anyone who wants to use this mapping.

Let me know what you think.

joebordes:
Look at this mapping, for example:
https://github.com/tsolucio/corebos/blob/master/modules/cbMap/processmap/Validations.php

it doesn't just process the XML, it actually validates all the values you give it.

joebordes:
That looks perfect.
I like the way the importCSV script is now just an example or template for all the functionality that is contained in the object.

Thanks !!!

Have a nice weekend

AlbanaCelepija:
I am glad to know it is ok now.
Thank you for the helping comments.
Have a nice weekend too !

PR #104, opened 2016-09-16
Better use of the language files when no records are found?

Luke1982:
This is a test that I'd like evaluated. The current behaviour leads to a very weird sentence in Dutch. I changed it so that it is more generic and makes sense in both English and Dutch ("No Records Found"). I know this would also work in German, but I have no idea how it would translate in Spanish, Italian etc.. Please let me know if this would be an improvement in these languages as well.


joebordes:
I think this is a good idea.
It doesn't solve the real problem which is the lack of a complete translating system which natively supports plurals but it does masquerade the fact in one of the places that it was really visible.
Following your lead I changed the whole section. I changed the message and added the import link to inventory modules.

https://github.com/tsolucio/corebos/commit/664e08ac0f94a456cc23ede628f961d1d64704db

Thanks!!

Luke1982:
Great, at least this solves the weird sentence structure (yes it does not solve the underlying issue but at least this is a lot better). You can update my theme now as well and see the latest changes there.

PR #114, opened 2016-11-02
(feat) some translations for account module (Dutch)

Luke1982:


modules/Accounts/language/nl_nl.lang.php
```diff
@@ -208,5 +208,12 @@
 	'Projects' => 'Projecten',
 	'Timecontrol' => 'Urenregistratie',
 	'InventoryDetails' => 'Voorraad Details',
+	'Exact VAT Code for Account' => 'BTW code voor klant',
+	'Is Converted From Lead' => 'Is gemaakt vanuit lead',
+	'Converted From Lead' => 'Gemaakt vanuit lead',
+	'Exact Payment Condition for Account' => 'Exact betalingsconditie voor klant',
+	'' => '#leeg#',
```
joebordes: Are these three correct. I think it will be better if you don't leave empty lines like that just in case.

joebordes:
I meant the empty lines, those didn't look correct. I got it, no problem.
Thanks!

PR #117, opened 2016-11-10
Divisible validation

Luke1982:
Creates validation for inventory modules, based on the "divisible" field in either the product or the service. Alerts the user when trying to buy/sell a product or service in quantities not divisible by one when this is set for that service or product. Translation strings are saved on products. Different translation string for purchaseorders since alerting about selling is not right there, so the string alerts about buying.


joebordes:
sorry for the delay

PR #118, opened 2016-11-10
Lightning Design evvtMenu

SkeLa-:


Luke1982:
I tried this. Is it supposed to only show the word "DEMO" in the topleft corner and not do anything else?

joebordes:
You have to apply the changeset, but you have to manually apply that change
It looks really clean and nice

Luke1982:
Ah, now I see it. Looks good!

joebordes:
Yes, I saw that one too and a few other things. SkeLa is going to send two more options, one based on KendoUI and another on jQuery. Once we see the three we can decide which one to adopt.
I am inclined to use the LDS system, I would like to implement that as a coreBOS standard, but let's see the other options first

Luke1982:
I like the idea of a role/department based menubar versus module-based. As soon as the choice is implemented I'll try and spend some time on my theme using these new techniques..

kapsule:
@joebordes. Que buena pinta tiene. Vamos a esperar el resto de opciones para ver cual se adapta mejor. Pero de momento la cosa pinta muy bien.

joebordes:
I ended up accepting this LDS menu

Thanks!!

PR #119, opened 2016-11-16
Kendo menu added

SkeLa-:
this branch has the kendoui menu, also it has permissions for menu elements.


joebordes:
I ended up accepting the LDS menu

Thanks!!

PR #123, opened 2016-12-02
Jquery menu

SkeLa-:
the jquery menu, has the same logic as the LD-menu, we can only change the html/css of the menu to a different style/theme


joebordes:
I ended up accepting the LDS menu

Thanks!!

PR #124, opened 2016-12-14
feat(Calendar4You) upgrade fullcalendar, gcal and remove Zend

loridacito:


joebordes:
It is working very well. Thank you :-)

I found one issue and have one comment

joebordes:
**Comment**:
We can now register links only for the module they are needed on or globally (which is the default).

https://github.com/tsolucio/corebos/blob/master/vtlib/Vtiger/Link.php#L112

Note the last parameter. I modified the changeset to load moment and the css only on the calendar (and I added the license)

https://github.com/tsolucio/corebos/blob/master/build/changeSets/addlinksforcalendar.php#L26

joebordes:
**Issue**:

The mini calendar is not working anymore. Before the update if you opened the mini calendar and clicked on a date the big calendar would go to that day. Now it sends you to January 1970.

calendarerrorgotofromminicalendar

joebordes:
Note: you can commit the fix directly to the master branch when you have it

PR #128, opened 2017-01-12
Open favorite Web page in other browser window

EduardoJimenez:
Many sites now do not allow to be embedded in iframes

### DIFF
--- a/Smarty/templates/MySitesContents.tpl
+++ b/Smarty/templates/MySitesContents.tpl
@@ -49,7 +49,7 @@
 </tr>
 <tr>
 	<td bgcolor="#ffffff" colspan=2>
-		<iframe id="locatesite" src="{$DEFAULT_URL}" frameborder="0" height="1100" scrolling="auto" width="100%"></iframe>
+		<!--<iframe id="locatesite" src="{$DEFAULT_URL}" frameborder="0" height="1100" scrolling="auto" width="100%"></iframe>-->
 	</td>
 </tr>
 </table>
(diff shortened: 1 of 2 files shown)


joebordes:
You have implemented an "all or nothing" solution.
What I had in mind was to detect if the website we are trying to load has X-Frame-Options set to deny and, in that case, show a message something like:

"The website you have indicated has blocked loading inside another webpage so you will have to open it in a new window/tab. Please click here if you want to access the site."

If the site is not blocked simply do what we do now.

That would require doing this:
https://stackoverflow.com/questions/21263418/detect-x-frame-options?rq=1

create a function to call this for the URL and act accordingly.

What do you think?

ecrmweb:
Good solution.

Thanks!

*Ranieri Slemer*


“CRM é uma estratégia para o crescimento e sobrevivência empresarial”.


2017-01-12 6:53 GMT-02:00 Joe Bordes <notifications@github.com>:

EduardoJimenez:
Can we safely assume that the curl module will be available in our installations?

If that is so, then I will include some logic based on the link given, and if it works, restore the iframe for the cases allowed.

Believe me, I tried to tap into the iframe properties to detect no page, but without any success.

joebordes:
Yes, you can count in having cURL installed. The application uses it in some places. We even have a high level interface for it:

vtlib/Vtiger/Net/Client.php: Vtiger_Net_Client

I can imagine, it happens to me every day :-)

Have a nice weekend

EduardoJimenez:
I have made a commit with the definitive fix, have you received it?

joebordes:
There are two PRs here about this issue. I see them both and will get around to testing and integrating them shortly.
I'll get back to you soon

PR #130, opened 2017-03-07
i18n(fr_fr) update FR translation

polanskiman:


joebordes:
We lost a few strings when I blissfully ignored the conflicts git was telling us about, but that was a calculated risk I had controlled as I didn't want to complicate it more for your first PR

I will accept the PR and then send you the list of changes so you can create a new branch and PR. that way you practice :-)

thank you VERY much for your hard work on this. I really appreciate it

joebordes:
these are the lost labels:

https://github.com/tsolucio/corebos/commit/c8f417e08ac56d940c5d73dc8d2c3e8cc9d42a7e

PR #147, opened 2017-05-05
(feat) Make "getRelatedRecords" return creator's first and last name

Luke1982:


joebordes:
Looks good.
How about doing the same for smownerid (owner_firstname and owner_lastname)?

Luke1982:
No problem, give me a sec.

Luke1982:
Tested and works for owner as well.

joebordes:
Perfect.
Just one comment on the commit message for the next ones you create: the format is detailed here http://corebos.org/documentation/doku.php?id=en:devel:commitguidelines

In this case it would be:

feat(Webservice) {message}

Thanks!!

Luke1982:
Ah yeah thanks, I knew that but forgot...

PR #151, opened 2017-05-27
feat(core): Allow module developers to set a custom listview template while retai…

Luke1982:
…ning all logic in Listview.php. You can set a flag like $custom_list_view = "modules/Module/template.tpl" in your module's Listview.php file. This way you can still use all the default data the Listview.php file produces and pushes to Smarty without reinventing the wheel.


joebordes:
Nice!
Thanks

PR #153, opened 2017-08-03
Enable returning to newly created record from related list ADD button

Luke1982:


Luke1982:
I've updated the PR. The only thing I doubt is [this](https://github.com/tsolucio/corebos/pull/153/commits/64941fe80e57467501b5f3494c4f40cc1a8ddcae#diff-1bac0e9dcb89bd82f4a6b47551aca0f6R126) part. I think this is wrong because it doesn't check if the property actually has a value.

joebordes:
I merged and tested this functionality.

I had to add a similar change for the source module value because it is also used and wasn't establishing the relation correctly in the crmentityrel table.

[This is my change](https://github.com/tsolucio/corebos/commit/5dc1364a001bf0b59e1d98a742d5aebd9bdae73b)

[I updated the documentation also](http://corebos.org/documentation/doku.php?id=en:devel:develtips#i_would_like_to_be_able_to_ret)

Thanks!!

PR #160, opened 2017-08-21
feat(GlobalSearch)autocomplete to launch a reduced global search

AlbanaCelepija:


joebordes:
It seems to be working correctly, there is just one issue that I would like to change. I create this issue to control it when you can.

#165

Thank you!

Issue #159, opened 2017-08-19
webservices/Update.php attachments issue


reetp:
Ok, may be it is me, and maybe not :-)

I am trying to add an Image to a Product but I think there may be an issue with the array handling in Update.php

here's the code:

```
	if (!empty($element['attachments'])) {
		foreach ($element['attachments'] as $fieldname => $attachment) {
			$filepath = $root_directory.'storage/'.$attachment['name'];
			file_put_contents($filepath, base64_decode($attachment['content']));
			$_FILES[$fieldname] = array(
				'name' => $attachment['name'],
				'type' => $attachment['type'],
				'tmp_name' => $filepath,
				'error' => 0,
				'size' => $attachment['size']
			);
		}
```
I realised this should theoretically handle multiple attachments.

I have an array as follows (see he screenshot):

$element
  => attachments
   ==> content
   ==> name
   ==> size
   ==> type

First issue is the $filepath... $attachment['name] shows the correct filename but I actually end up with a path like this:

....corebos/storage/i

So the $filepath seems not quite right somehow.

file_put_contents then fails. 

Second I would have expected it to have been something like:

.....corebos/storage/2017/August/weekx/123456_image001.jpg ??(where 123456 is the id)

![attachments1](https://user-images.githubusercontent.com/3800384/29488133-cba50972-8504-11e7-89b7-571cab5d2bd7.png)

joebordes:
First, this code you are looking at is for normal image fields. In other words, you create various image fields on a module (uitype 69), each is an individual image field with its own name. You can then upload images to any of those fields with the array structure you are looking at.

The special multi-image field on products (uitype 69m) is **NOT** supported by webservice.

The 
```
$filepath = $root_directory.'storage/'.$attachment['name'];
```
is a cache, temporary upload, which, in fact, has been moved to the cache directory:

https://github.com/tsolucio/corebos/blob/master/include/Webservices/Update.php#L39

the "save" process will copy them from this temporary cache to the definite storage directory.

HTH

reetp:
Oh bother.........

Is it possible to create just the link with webservices (on the basis I could copy image files across manually) ?

I can retrive the image link info ok with webservices.

joebordes:
I understand that you want to upload a set of images to the mulit-image field in your products.

We added to coreBOS a mass upload feature a few years ago:

https://youtu.be/usWQDoCVw_w

I had a look at that and it uses a service end point we added specifically for that use:

[...]

PR #167, opened 2017-08-23
New WS Product Image uploader

reetp:
Here is some sample code for the Product uploader as per:

https://github.com/tsolucio/corebos/issues/159

I'm pretty sure that uploadProductImagesWS.php is OK.

But there are gaping holes in UploadProductImages.php due to my lack of knowledge. It does actually work.

I think the file location are correct. The registration file needs copying to wsChanges and InstallRESTCHanges.php needs running to add it in to the DB.

Once you have stopped giggling, let me know :-)

### DIFF
--- a/build/wsChanges/uploadProductImagesWS.php
+++ b/build/wsChanges/uploadProductImagesWS.php
@@ -1,0 +1,26 @@
+<?php
+/*********************************************************************************
+ * Copyright 2016 JPL TSolucio, S.L.  --  This file is a part of coreBOS.
+ * You can copy, adapt and distribute the work under the "Attribution-NonCommercial-ShareAlike"
+ * Vizsage Public License (the "License"). You may not use this file except in compliance with the
+ * License. Roughly speaking, non-commercial users may share and modify this code, but must give credit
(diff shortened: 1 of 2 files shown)


joebordes:
Hi John

I reviewed and restructured the functionality. You were very close, good work :-)
These are the main changes I made:

- [commit guideline](http://corebos.org/documentation/doku.php?id=en:devel:commitguidelines). I would have put:
```
feat(Webservice) add support for mass upload of images into product multi-image field
```
- security first. the lines from 21 to 50, which you had at the end, make sure that we are working with the products module, with a valid product ID and that the user connecting has permission to read and write on that module
- lines 51 to 62 check if the current products record already has the maximum number of images
- lines 69 to 89 upload the images. important here is:
 - the basename in line 73 to make sure  we avoid path traversal issues
 - I changed the input array to be consistent with the other upload image webservice endpoints
 - the condition in line 86 that stops uploading images if we reach the maximum
- lines 90 to 97 return the result

[This is the commit](https://github.com/tsolucio/corebos/commit/76687b8f6e5f978dc12bd649da583817796b948c)


Thank you very much :-)
Have a nice weekend

PR #171, opened 2017-09-01
feat (WebServices) Add more detail to UploadProductImages return statement

reetp:
I added some code to the return statement to aid error checking/debugging

This helps when adding multiple images. If you exceed the file limit it will show which files have been added, and an error code.


joebordes:
Looks good. Have you launched the test script in coreBOSWSDevel?

reetp:
Not yet.... I was trying to resolve this first :-)

I'll try and do so this week.

PR #172, opened 2017-09-12
Multiple problems with variable and reporting detected

kapsule:
When running the Workflow planner PHP generated the following error.

PHP Fatal error:  Cannot use [] for reading in /var/www/parkare/modules/Emails/class.phpmailer.php on line 935
Fatal error: Cannot use [] for reading in /var/www/parkare/modules/Emails/class.phpmailer.php on line 935


kapsule:
Fixed new bug with the generation of reports.

joebordes:
I manually merged this one because I want to follow a slightly different block syntax and I decided to fix the array assignment error by updating PHPMailer to their latest version where they fixed it already.

Thank you VERY much !! :-)

PR #174, opened 2017-09-18
feat(Products) Load images in related products

kapsule:
Load images in related products


joebordes:
Nice feature. Added it to the application.

Thanks!!

PR #179, opened 2017-09-21
fix(Reports) adding new column for report type

albadisha:


joebordes:
It seems to be working correctly.
A few comments:
 - really nice of you to clean up the vtcache file :-)
 - next time separate the clean up from the fix, that makes it easier to see what has changed with less noise
 - I fixed a couple of warnings that you introduced, try developing with error reporting set to E_ALL
 - I see you found the commit guide lines :-)

Thank you very much!

albadisha:
I'm glad :)                                                                                                                                                        
I'll be aware of what you said, if I commit again later on.                                                                                  Thank you :)

Issue #197, opened 2017-10-02
Change name from Trouble Tickets to Support Tickets


joebordes:
Change the helpdesk module name

jsjqian:
Hi! I'm currently working on this issue and had a question about the name change. I changed the module name for the en_us and en_gb languages but I was wondering if the scope of this issue includes all other languages in the directory as well. Please let me know and I can change the names for the other languages as well. Thanks!

joebordes:
In general, I try to change as many language files as I can, but my knowledge is limited so I do trust the other language translators to follow the en_us change if they see fit.

I would say, change those that you are fluent in the language. Leave any that you would need google translator :-)

Thank you VERY much for your help!

jsjqian:
Thank you for the quick response! The only languages that I am fluent in of the ones in the directory are the ones that I have already changed. Looking through the other languages, I noticed that some only use "Tickets" instead of a translated version of "Trouble Tickets". I don't know if that's intended or if it's because I am not very familiar with the other languages but I just wanted to point it out if it was a mistake.

I have made a pull request at #208. Let me know if it needs any changes!

joebordes:
Hi,

There are more labels than the ones in the two files you changed. I launched a search for "Trouble tickets " in the whole application and about 60 relevant lines came up. Can you please have a look at all of those? If not, let me know.

Thanks!

joebordes:
Let me know if you have doubts about any change.

jsjqian:
I'll take a look and get back to you!

jsjqian:
Hey so I looked through the instances where "Trouble" appeared in the entire project. I only made modifications to the actual language files and in those only to the actual value and not the key itself. I saw that the database uses strings like "Top Trouble Tickets" to read and write and I felt like those shouldn't be changed at least for now. Also, any comments that used "Trouble" was left as "Trouble" because much of the code refers to the HelpDesk/Tickets as "Trouble Tickets" so I thought that by only changing the visualized values that it would cause as little confusion as possible. Let me know if that was what you had wanted and if not, what other parts I should also look at to resolve the issue. Thanks!

joebordes:
Closed via #208
Thanks!

PR #208, opened 2017-10-02
Switched 'Trouble' to 'Support' in en_gb and en_us

jsjqian:
Changed the module name in the en_gb and en_us files. Also swapped "Trouble Tickets ID" for "Support Tickets ID" for consistency. #197


joebordes:
Thank you very much!! :-)

PR #210, opened 2017-10-06
The calendar.js changes

### LINKED ISSUE #209: calendar.js unreachable code after return statement
I noticed that Firefox complains with Vtiger in the console
calendarvtiger

I had a look at the coreBOS demo and it has the same issue (though on a different line it's the same code)
calendarcorebos

coreboscode
It looks like everything after the first return is unreachable.

It looks like it could be that 

	if (!Calendar.is_ie5)
		//	return str.replace(re, function (par) { return s[par] || par; });
		// fix provided by dtnadmin. Many thanks
		for (var i in s) str=str.replace(i, s[i]);
		return str;

isn't closed properly and should be something like 

	if (!Calendar.is_ie5) {
		//	return str.replace(re, function (par) { return s[par] || par; });
		// fix provided by dtnadmin. Many thanks
		for (var i in s) str=str.replace(i, s[i]);
		return str;
	}

Making that change with a test Vtiger install stops FF from complaining. So the diff for coreBOS would be (assuming that's the right fix)

	--- a/jscalendar/calendar.js
	+++ b/jscalendar/calendar.js
	@@ -1693,12 +1693,12 @@ Date.prototype.print = function (str) {
	 	s["%%"] = "%";		// a literal '%' character
	 
	 	var re = /%./g;
	-	if (!Calendar.is_ie5)
	-
	+	if (!Calendar.is_ie5) {
	 		//	return str.replace(re, function (par) { return s[par] || par; });
	 		// fix provided by dtnadmin. Many thanks
	 		for (var i in s) str=str.replace(i, s[i]);
	 		return str;
	+	}
	 
	 	var a = str.match(re);
	 	for (var i = 0; i < a.length; i++) {

thanks :)

2038AD:
Here's that PR :) should resolve #209

### DIFF
--- a/jscalendar/calendar.js
+++ b/jscalendar/calendar.js
@@ -81,8 +81,6 @@ Calendar._C = null;
 Calendar.is_ie = ( /msie/i.test(navigator.userAgent) &&
 		   !/opera/i.test(navigator.userAgent) );
 
-Calendar.is_ie5 = ( Calendar.is_ie && /msie 5\.0/i.test(navigator.userAgent) );
-
 /// detect Opera browser
 Calendar.is_opera = /opera/i.test(navigator.userAgent);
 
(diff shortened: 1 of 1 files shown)


joebordes:
You cleaned the references to the ie5 variable. Nice !! :-)

Thanks!

PR #211, opened 2017-10-11
 feat(Calendar4You) clear tokens button when changing gmail account for the same user

loridacito:


joebordes:
Looks good. Thanks :-)

PR #212, opened 2017-10-12
Product image proportions

2038AD:
I've changed DetailViewUtils.php so that it no longer sets images over a certain size to be 450 by 300 (changing the proportions). It now sets a max-width and max-height so images are still limited in size but the proportions should stay the same so images don't appear squashed. Thanks :)


joebordes:
Looks perfect!  Thanks :-)

PR #214, opened 2017-11-16
Denormfields listview

evolutivo:


joebordes:
this has been implemented taking another (more broder) approach on branch denomarlize_module. development is entering testing phase

PR #226, opened 2018-01-18
feat(Settings) Implemented improvement for the calculation of permiss…

kapsule:
…ions per profile and code format.


include/utils/UserInfoUtil.php
```diff
@@ -3790,12 +3790,19 @@ function getPermittedModuleIdList() {
 /** Function to recalculate the Sharing Rules for all the users
   * This function will recalculate all the sharing rules for all the users in the Organization and will write them in flat files
  */
-function RecalculateSharingRules() {
+function RecalculateSharingRules($roleId) {
```
joebordes: We have to make this parameter optional:

```
function RecalculateSharingRules($roleId = 0) {
```
because it is currently called in a dozen places that would start emitting warnings if not

modules/Settings/CalculatePrivilegeFilesRole.php
```diff
@@ -0,0 +1,19 @@
...
+ * are made available under the terms of the GNU Lesser General Public License
+ * (LGPL) version 2.1 which accompanies this distribution, and is available at
+ * http://www.gnu.org/licenses/lgpl-2.1.html
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+*/
+$roleid = vtlib_purify($_REQUEST['roleid']);
```
joebordes: Please add admin only access protection here:

```
global $current_user;
if (!is_admin($current_user)) {
....
```

just for security

joebordes:
Looks nice!!  Thank you!

I added a couple of changes I need you to add and I would like to add one more comment which is to separate the code formatting changes in the future if possible. I like commits to be cohesive, mostly related to the change they implement.

In any case, the clean up is perfect, it is preferable to have it than not

Let me know if you have any doubts about the changes I requested.

modules/Settings/CalculatePrivilegeFilesRole.php
kapsule: Hi @joe,

Perfect, thank you very much for your comments. As always, it is a pleasure to learn from a great master.

Regards

kapsule:
@joebordes, ya he subido los cambios que has propuesto. Saludos.

joebordes:
Perfect!  Thanks!

Have a nice weekend

PR #227, opened 2018-01-19
fix(Users) Improve permissions calculation performance by introducing multiple value insert in tables instead of single value

afrohoxha:


joebordes:
Nice catch. On my development install which has a very small set of users and all with the same configuration more or less I measured a benefit of 0.05s (!)

Thanks!!

Note that you are working with a VERY old coreBOS, we have made a lot of enhancements in all areas, you should consider updating.

afrohoxha:
Well, I made tests with Subito installation and since it has 500+ users, the calculation was really slow. Precisely, db inserts finished after several hours even in production environment. In demo environment it took almost 10+ hours to finish with single insert statements and 15 mins with multiple values insert ( calculation for all users). I agree with you, for small number of users and modules, it has unnoticeable difference and no benefits since the size of the table is too small to have considerable performance degradation because of index inserts and foreign key checks.

PR #228, opened 2018-01-21
feat(Users) Implemented option to export user list

kapsule:


joebordes:
Looks nice!

Changes that I need to accept the PR:

1.- include/utils/ExportUtils.php

You have reverted the formatting in this file. Please undo your formatting changes and leave only the user fields "if"

2.- modules/Users/Users.php

Please apply coreBOS Standard code format to the new function. Although the rest of the file is not formatted yet at least new code we add should be in the correct format.

3.- modules/Users/ExportRecords.php

Copy this file from here: https://github.com/tsolucio/corebos/blob/master/vtlib/ModuleDir/ExportRecords.php

the VPL license obligates us to do that.

4.- modules/Users/ExportRecords.php

As in the previous change, add an "if (is_admin($current_user))" so that only admin users will be able to get the CSV.

Thanks!!!

joebordes:
Thanks!!

PR #241, opened 2018-02-12
 feat(LayoutEditor) Check field usage before delete

loridacito:


joebordes:
I manually merged this PR.

I modified the LDS modal as the footer was off and the buttons weren't styled.

I added the Lead Conversion check

PR #243, opened 2018-02-20
Check field usage load data and translation

evolutivo:


evolutivo:
This commit is for the two remaining issues of layout editor. The formatting of the code has been added to the same commit (sorry!). 
The only changes made are the addition of translation strings in modules/Settings/language files starting from this comment.
//loaddata language translation strings

, the usage of the $mod_strings variable in loaddata.php

and also we checked queries and using like is correct because the fieldname is not fixed in table columns e.g  in cvcolumnlist: vtiger_leaddetails:lead_no:lead_no:Leads_Lead_No:V
It has been added a condition to check the module in loaddata.php:
if (!empty($fldname) and !empty($modname)) { $mod = Vtiger_Module::getInstance($modname); if ($mod) { $field = Vtiger_Field::getInstance($fldname,$mod); if ($field) {....

joebordes:
I'm sorry but I could not accept this PR.

- You were applying incorrect formatting, please read the wiki page
- You were using an old version of the files, I had already applied some formatting and changes
- You mixed formatting and logic. Next time remember that the git "-p" parameter is your best friend

In any case, I manually applied the changes and closed the issues. Thank you very much for your help!!

PR #244, opened 2018-02-23
bug on external reports

denaldd:

PR #245, opened 2018-02-27
feat(Reports) external reports new tab

evolutivo:


joebordes:
@denaldd, can you please inform me of what you are fixing. I follow the steps on the #244 issue with no success. The report opens incorrectly inside the window. It doesn't open the URL I selected.

Look for me on hangouts and we can do a screen share so I can understand what you are doing and I can show you what I am doing (and expecting).

PR #246, opened 2018-02-28
feat(Reports) open report in new tab

evolutivo:


evolutivo:
Sorry @joebordes , the correct pull request is this one. We had misunderstood the request. In this one the external link opens in a new tab

joebordes:
Ok, now it seems to be solving the issue in the first part of the problem.

Now the second part is still pending. When you open a report in the same folder of the external report, you will find in the upper right corner a picklist which permits you to load another report. We must modify that picklist to open the external report in another tab also.

Thanks!!

PR #247, opened 2018-02-28
feat(Reports) open report selected from picklist  in new tab

evolutivo:


joebordes:
Close but not correct. The url returned from the application in selectReport is the application report url, not the url of the external report:

```
"index.php?action=SaveAndRun&module=Reports&record=47&folderid=1"
```
so it does open in another window but it opens the incorrect content.

Additionally, in the formatting commit use the message:

chore(Reports) cbSR
cbSR(Reports)
style(Reports) cbSR

all of those commit message make it really easy to simply ignore them when you are looking for some particular change. The message "feat(Reports)opening report to new tab" implies that I have to look inside to see what has changed.

In any case, thanks for taking the time to apply formatting!

PR #250, opened 2018-03-11
Inventory details tax block

Luke1982:
This PR installs two new tax events:
- corebos.changestatus.tax
- corebos.changelabel.tax

On top of that, it installs a 'Tax' block in the InventoryDetails module that holds all system inventory taxes. It then installs the existing taxes as fields in the new Tax block.

Then it registers some event handlers that updates the appropriate fields in the new taxblock when taxes are created, disabled of enabled and when the label changes.


Luke1982:
One issue I can think of is if the column type for the percentages are correct: https://github.com/tsolucio/corebos/pull/250/files#diff-255a59c2ec9fc25545c54fb6171a4387R40

joebordes:
Percentage fields are defined as decimal(7,3) so you will have to change those field definitions.

Also, all event handlers are called on every save, which is why we have to add extra "if" protection on all handler methods. In this case, you will have to add a condition to detect the correct event you need besides the 'tax_type' index.

You can find some [more information here](http://corebos.org/documentation/doku.php?id=en:devel:corebos_hooks#event_launching)

Finally, if you can add the new block label to the InventoryDetails language files that would be nice. If not, it's ok, I'll do it.

Besides that, it looks good.
Thanks!!

Luke1982:
I fixed the column type and also committed all translations (as far as my Google-language is correct, please correct where I messed up).

As far as the if clause for the modules checking, I'm not sure if I agree there. I know that you're supposed to use that format when registering handlers that listen to records being saved, deleted and such. But these handlers don't fire on every record, or even on any record for that matter. They fire on adding a tax, updating a tax label and enabling/disabling a tax. So there really isn't a module to check on (besides 'Settings' maybe). I didn't configure the [events](https://github.com/tsolucio/corebos/pull/250/commits/592676081e8c188dbabd7f7a1ecb867bc869d3cb) to fire on record editing/saving. Or am I wrong here and do ALL events fire on modules ALWAYS, even the ones I created, even though they don't live in module-code? In which case, should I check on the module name 'Settings'?

joebordes:
You are right, those events will only be called directly. Sorry about that.

PR #251, opened 2018-03-13
Inventory details tax block

Luke1982:
Sorry, started a new PR becuase the old one was getting messy and cluttered with typo-commits.

As far as the if clause for the modules checking, I'm not sure if I agree there. I know that you're supposed to use that format when registering handlers that listen to records being saved, deleted and such. But these handlers don't fire on every record, or even on any record for that matter. They fire on adding a tax, updating a tax label and enabling/disabling a tax. So there really isn't a module to check on (besides 'Settings' maybe). I didn't configure the events to fire on record editing/saving. Or am I wrong here and do ALL events fire on modules ALWAYS, even the ones I created, even though they don't live in module-code? In which case, should I check on the module name 'Settings'?

### DIFF
--- a/build/changeSets/2018/inventoryDetailsTaxBlock.php
+++ b/build/changeSets/2018/inventoryDetailsTaxBlock.php
@@ -1,0 +1,102 @@
+<?php
+/*************************************************************************************************
+ * Copyright 2018 MajorLabel -- This file is a part of MajorLabel coreBOS Customizations.
+* Licensed under the vtiger CRM Public License Version 1.1 (the "License"); you may not use this
+* file except in compliance with the License. You can redistribute it and/or modify it
+* under the terms of the License. MajorLabel reserves all rights not expressly
+* granted by the License. coreBOS distributed by MajorLabel S.L. is distributed in
+* the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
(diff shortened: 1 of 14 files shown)


joebordes:
You are right, those events are only called directly. Sorry about that.

joebordes:
Ok, it looks good. I will accept it and make a few adjustments to formatting. Mostly whitespace and the Array() which is array. There should not be any captilizations in PHP language.

I have a question: where are you filling in these tax fields? Shouldn't there be a change in the InventoryDetails save method?

joebordes:
In case you want to [see the formatting](https://github.com/tsolucio/corebos/commit/012090b4db4923d2fa9f6ff1e2cf76533b07c04c)

Luke1982:
Ah, I was under the assumption that capitalization was more a pre than a con, but it's the other way around. Anyway I didn't do all the caps, I think my IDE did some of those (sublime). Also, I didn't know require functions were preferred without braces. Will keep that in mind.

About the filling, this basically is a part of my InventoryBlock project (idlines). We could alter the save method that works on the 'old' product lines to fill out these fields, but if we're moving towards a replacing that with my project, I'd rather focus on that. FIlling would be done through a webservice call when done via AJAX (like you proposed) or though a normal save call (where you would fill out the column_fields array) when using an aftersave of beforesave event on the master module.

joebordes:
ok, got it. I am also keen on moving forward with your project

Issue #199, opened 2017-10-02
Filter conditions "does not start/end with" for filters and workflows


joebordes:
Add the two missing filter conditions:

- does not start with
- does not end with

PR #253, opened 2018-03-22
feat(Filter conditions does not start/end with for filters and workflows

evolutivo:
Issue: https://github.com/tsolucio/corebos/issues/199

There is one more commit in the code due to the last merge which was made manually. Can you accept the PR and remove those changes or should I create a branch in my repo starting from the master of tsolucio/corebos?


joebordes:
Thanks!

PR #254, opened 2018-04-05
Updates to LDAP

reetp:
This fixes a few typos, fixes a function name, adds Street and City fields, adds a function to generate a more complex default password if required


reetp:
Sorry - not sure how I leave out those pull requests where we merged yours back to mine !!!

joebordes:
to avoid the extra merge and commits you must always create a new branch and make your changes there. Normally something like this:

```
git checkout master
git pull upstream master
git checkout -b my_awesome_pr
add your changes and commit here it will go to the branch 
git push
git checkout master
wait for PR merge
git pull upstream master
you can delete your branch now as the changes are in master
```

HTH

Thanks!!

reetp:
Thanks Joe ! My git skills are pants :-(
I think it would be better to add some of the vars as Global Config Vars and remove some or all of config.ldap.php (which should really be in the root dir with the other config files). Maybe I can revisit it when I have nothing better to do :-)
I think this all works OK but obviously happy to look at any issues found.

PR #255, opened 2018-04-15
fix(GContacts)handle synch exception,clean up code fix #220

### LINKED ISSUE #220: Google Contacts sync errors
I have found two errors in the Google Contacts Sync code:

https://github.com/tsolucio/corebos/blob/master/modules/Contacts/connectors/Oauth2.php#L140
Neither VtigerConfig nor OAUTHREDIR_PUBK exists

https://github.com/tsolucio/corebos/blob/master/modules/Contacts/views/List.php#L65
Zend_Gdata_App_HttpException does not exist

AlbanaCelepija:


joebordes:
Hi, nice to see you around again :-)  Hope all is well.

This change looks good but I noticed that there is no service column in the its4you_googlesync4you_access table. I am going to accept this and open a new ticket with the issue so you can have a look what is missing.

Thanks!!

AlbanaCelepija:
Hi Joe,
Hope all is well too !
Regarding the table its4you_googlesync4you_access it actually has one column service and i didn't see any error on logs when removing sync or even when logging out the email account of google contact.
Tell me if it is necessary anything else to reproduce the issue.
Regards

joebordes:
Let's continue the conversation in issue #256

PR #257, opened 2018-04-16
Update LayoutBlockList.php

besartmarku:
Showing 'field size' too at the layout editor ->properties (when admin edits a custom field)


joebordes:
Thanks!!

PR #258, opened 2018-04-16
Update LayoutBlockEntries.tpl

besartmarku:
-removing column name from table header 
-creating new row at first table to show "column name","type" and "fieldsize"


joebordes:
merged in #257

PR #261, opened 2018-04-19
GCalendar fix #256

### LINKED ISSUE #256: Google Contacts missing column
On these lines (at least):

https://github.com/tsolucio/corebos/blob/master/modules/Contacts/models/Module.php#L15
https://github.com/tsolucio/corebos/blob/master/modules/Contacts/models/Module.php#L32

there is a reference to a column that does not exist: its4you_googlesync4you_access.service

AlbanaCelepija:


joebordes:
Hi

I see you added the service column to Calendar Sync. You added it to the manifest which is good but you have to add also a changeset that will create (and fill) the field for existing installs, those that will never execute the manifest but that may have the service already configured.

At the same time, I suppose you can delete this line:

https://github.com/tsolucio/corebos/blob/master/include/integrations/GContacts/GContacts.php#L84

as it won't be necessary anymore.

Ok?

PR #262, opened 2018-04-20
feat (Profile) edit using ajax

loridacito:


joebordes:
I read the code and did some tests but it is not clear to me if you have reduced the number of variables or not.

I understand that you have changed the form post for an ajax post, but it seems to me that the number of variables sent is still the same.

I put a ``` count($_REQUEST); ``` in UpdateProfileChanges.php and the number of variables that get there are the same so we still have the input_max_vars limitation. Or am I missing something?

joebordes:
Thanks for the formatting :-)

loridacito:
You are right. I got too much  into changing the form submit for an ajax call and I lost sight of the real reason we were doing the change , sorry. I will make the changes on monday

joebordes:
no problem :-)
Thanks

joebordes:
Nice !!  :-D

Thank you!

PR #274, opened 2018-05-22
Fixing mandatory fields to display Fields Labels

tebajanga:
A fix to display mandatory fields labels on errors.


joebordes:
Thanks!

PR #275, opened 2018-05-24
feat(MassEdit) add worker for running mass edit

loridacito:
As I told you the EventSource and even the Worker doesn't block the UI but doesn't let you change pages either. Another page chosen from the menu keeps loading until the mass edit has reached 100%

### DIFF
--- a/Smarty/templates/ListView.tpl
+++ b/Smarty/templates/ListView.tpl
@@ -138,6 +138,36 @@
 <div id="massedit_form_div"></div>
 
 </div>
+<div id="relresultssection" style="visibility:hidden;display:none;" class="slds-masseditprogress">
+<div class="slds-grid">
+<div class="slds-col">
+	<div class="slds-page-header" role="banner">
+		<div class="slds-col slds-has-flexi-truncate">
+			<div class="slds-media slds-no-space slds-grow">
+				<div class="slds-media__figure">
+					<svg aria-hidden="true" class="slds-icon slds-icon-standard-user">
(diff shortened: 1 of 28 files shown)


loridacito:
added server validation in masseditsave

joebordes:
Looks good. Just one comment: on line 84 of MassEditSave you call this:

```
$exists = executefunctionsvalidate('ValidationExists',$currentModule);
```

that call is inside the loop but only depends on the module name so you can move this call outside the foreach to save some time and memory.

loridacito:
Yes I didn't notice that . In fact I was thinking that even the second validate function doesn't depend on the loop. I can move that out too

loridacito:
Although I think there is smth 'wrong' with the mass edit validation. For example I added a validation in Campaigns module checking that campaignname should be different from campaigntype. It checks it correctly with the value of the campaintype in massedit,  --None-- if it is not selected,  and not with the value of the field in the record. But it used to work like this even before

joebordes:
Yes, currently that is how it works. We do not have access to the values on the record, you must validate the values that are coming from the user or explicitly program the validation to go get the value in the database. At least for the moment because I have that on the immediate roadmap.

Looking at the code now that you have extracted the validation I would propose one of these changes:

- if the validation fails, it will fail for all records so just send one message and finish, so as to not send many times the same message repeated
- move the validation back into the loop so when we actually add the possibility to use the existing database values it will work correctly for each record: some will be saved and others won't depending on the validation

loridacito:
Ok I put the second validation function back in the foreach

joebordes:
Thanks!!

PR #280, opened 2018-06-08
enhance convert inventory module task to support all modules with related products/services

Mikel1995:
- Modify convert_inventorymodule.inc script 

- Create changeset for allow task to appear and in non inventory modules


joebordes:
thanks!

PR #283, opened 2018-06-11
Security(filter) XSS showing description on list view.

orionliciAT:
Security(filter) XSS showing description on list view.


joebordes:
I am closing this change in favor of a more specific one that applies only to the field type with the problem (uitype 19 and maybe 20)

PR #287, opened 2018-06-21
CoreBOS Console

albertxhani:
@joebordes  this is the corebos console with some of the commands I thought were very useful.
In order to test it in your local machine, you have to run
`composer install` in the root directory
and then `php console`


joebordes:
Nice!

We need to document this in the wiki. Can you dedicate some time to that?

PR #291, opened 2018-06-29
Fix lead conversion mapping error message

Mikel1995:


joebordes:
Remember to answer in the forum with a link to the commit.

Thanks!

PR #294, opened 2018-07-04
feat(Workflows) Compare time fields as a qty of seconds and send e-ma…

Luke1982:
…ils using the direct field values. 

With this commit you can compare two uitype 14 fields with each other as quantities of time (e.g.: is 12:23:45 a longer time than 04:34:12?). This allows you to send e-mails from the workflow task that warns a user about too little / too much time spent (for example) and to mention the quantities of time in question.


joebordes:
Thanks!

PR #296, opened 2018-07-05
Aggregate time

Luke1982:
…ormat.

We could also include the operation into the function call?


joebordes:
Looks good to me. I like the way you reuse __cb_aggregation_operation :+1:

PR #298, opened 2018-07-05
feat(InventoryModules) Add a GV that allows you to turn off the defau…

Luke1982:
…lt PDF links

This is useful when you use GenDOC of PDFmaker (or any other alternative document creator) to hide the links you don't use anyway.


joebordes:
I am going to close this PR without accepting it because we are currently working on a project to move all actions to a configuration module. That would make this GV obsolete so I prefer to hold on a little longer.

Luke1982:
Cool, I'll wait for that then and continue with a tiny hack-comment for now.

PR #309, opened 2018-07-12
feat(Basic/Advanced Search) Translate picklist values for searching

loridacito:


joebordes:
Hi
This change looks really nice: small and effective :-)

A couple of comments:

1.- The first test I launched, did not work because I defined the translations for the picklist using the "For Picklist" field available on the translation module, not the "For Field" field. I made this change and it started working:

```
diff --git a/include/QueryGenerator/QueryGenerator.php b/include/QueryGenerator/QueryGenerator.php
index 2cba75344..b1d019a14 100644
--- a/include/QueryGenerator/QueryGenerator.php
+++ b/include/QueryGenerator/QueryGenerator.php
@@ -967,7 +967,7 @@ class QueryGenerator {
                                                $fieldSql .= "$fieldGlue DATE_FORMAT(".$field->getTableName().'.'.$field->getColumnName().",'%m%d') ".$valueSql;
                                        } else {
                                                if ($field->getUIType() == 15 || $field->getUIType() == 16) {
-                                                       $fieldSql .= "$fieldGlue ".$field->getTableName().'.'.$field->getColumnName().' IN (select translation_key from vtiger_cbtranslation where locale="'.$current_user->language.'" and forfield="'.$field->getColumnName().'" and i18n '.$valueSql.')'
+                                                       $fieldSql .= "$fieldGlue ".$field->getTableName().'.'.$field->getColumnName().' IN (select translation_key from vtiger_cbtranslation where locale="'.$current_user->language.'" and forpicklist="'.$this->getModule().'::'.$field->getColumnName().'" and i18n '.$valueSql.')'
                                                                        . ' OR '.$field->getTableName().'.'.$field->getColumnName().' '.$valueSql;
                                                } else {
                                                        $fieldSql .= "$fieldGlue ".$field->getTableName().'.'.$field->getColumnName().' '.$valueSql;
```
I created the "for picklist" field exclusively for this case, but the "for field" also can be used. So the question is which one do we use?  I would prefer the "For Picklist" because the other one is for translating the contents of a field.

2.- The global, unified search is not supported. Can we get that working also?


Thanks!!

joebordes:
Thinking about this change two more ideas came to mind:

3.- Currently we try to transform the search string in to the original value, so if you search for "Cerrado Ganada", internally we will actually search for "Closed Won", this transformation has a high cost and produces false positives in some cases, so I was thinking that we should eliminate that transformation now that we can actually search correctly. What do you think?

4.- Shouldn't we create a changeset to import the picklist translations? Something like [this one](https://github.com/tsolucio/corebos/blob/master/modules/cbtranslation/move2db.php) but just for picklists:

evolutivo:
Hi Joe, 

1. Ok we should use forpicklist then
2. I have to check it but I think we can
3. I don't understand what you mean by : internally we are searching for Closed Won.  Is it for the fact that we use the $field->getColumnName() IN (select translation_key..) in the query ?
4. Ok I can create a changeset

joebordes:
3.- Currently, when a string is given to search in a picklist, the application tries to translate it to the value in the picklist. Look at the image below, where you can see that introducing the text "Cerrada Ganada" it actually finds the English version and uses that in the SQL to find the right records.

plsearch01

This happens in the ```getTranslationKeyFromTranslatedValue``` [function here](https://github.com/tsolucio/corebos/blob/master/include/utils/utils.php#L2253), but that produces false positives from time to time because we can't really be sure we are getting the right original string from the translation file.

Also, this function has a high cost in memory and processing and I think it is not needed anymore with the change you are implementing here, so I am asking if you can confirm that this method is no longer needed and deprecate that function and eliminate it wherever it is being used.

Keep asking if it is still not clear.

evolutivo:
The first three points are done

PR #310, opened 2018-07-13
feat(cbMap) Allow multiple validation of the same type on one field

Luke1982:
Previously, if you would apply multiple rules of the same type (e.g. 'custom' on the same field), only the last one would apply. With these changes you could use more than one (especially helpful when using custom rules) validation of the same type on one field.


joebordes:
This change looks good to me. I'll try to keep an eye on it for a few weeks just in case, but I'd say that is correct.

Thanks!

PR #311, opened 2018-07-13
Create getRelatedModules Webservice call

Mikel1995:


build/wsChanges/CPgetRelatedModules.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/*********************************************************************************
+ * Copyright 2012-2014 JPL TSolucio, S.L.  --  This file is a part of coreBOSCP.
```
joebordes: set year to 2018

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,51 @@
+<?php
+/***********************************************************************************
+ * Copyright 2012-2014 JPL TSolucio, S.L.  --  This file is a part of coreBOSCP.
```
joebordes: set year to 2018

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,51 @@
...
+ * on an  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the
+ * License terms of Creative Commons Attribution-NonCommercial-ShareAlike 3.0 (the License).
+ ************************************************************************************/
+require_once 'include/database/PearDatabase.php';
+require_once 'include/ComboUtil.php';
+require_once 'include/utils/CommonUtils.php';
+require_once 'vtlib/Vtiger/Language.php';
+require_once 'modules/PickList/PickListUtils.php';
+
```
joebordes: I don't think you need any of these requires. Once you have the unit test working and passing, eliminate them and make sure the test still pass. Then, just in case, use the coreBOSWSDevel tool to call the service and see if it works.

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,51 @@
...
+		$rel_tab_id = $adb->query_result($result, $i, 'related_tabid');
+		//$function_name = $adb->query_result($result, $i, 'name');
+		$label = $adb->query_result($result, $i, 'label');
+		$actions = $adb->query_result($result, $i, 'actions');
+		$relationId = $adb->query_result($result, $i, 'relation_id');
+		if ($rel_tab_id != 0) {
+			if ($is_admin || $profileTabsPermission[$rel_tab_id] == 0) {
+				if ($is_admin || $profileActionPermission[$rel_tab_id][3] == 0) {
+					// vtlib customization: Send more information (from module, related module) to the callee
+					$focus_list[$label] = array('related_tabid' => $rel_tab_id, 'label'=> $label, 'labeli18n' =>getTranslatedString('SINGLE_'.$label, $label), 'actions' => $actions, 'relationId' => $relationId);
```
joebordes: You cannot guarantee that the label will be the module name so this is incorrect ```getTranslatedString('SINGLE_'.$label, $label)``` you have to use

```
getTranslatedString($label, getTabModuleName($rel_tab_id))
```

The same applies to line 46

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and limitations under the
+ * License terms of Creative Commons Attribution-NonCommercial-ShareAlike 3.0 (the License).
+ ************************************************************************************/
+require_once 'include/database/PearDatabase.php';
+require_once 'include/ComboUtil.php';
+require_once 'include/utils/CommonUtils.php';
+require_once 'vtlib/Vtiger/Language.php';
+require_once 'modules/PickList/PickListUtils.php';
+
+function getRelatedModulesInfomation($module) {
```
joebordes: All web service methods get passed in, as a last parameter, a user object that represents the user making the call, so you must add an additional parameter called ```$user```

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,51 @@
...
+	global $log, $adb, $current_user;
+	$log->debug('Entering getRelatedModulesInfomation(' . $module . ') method ...');
+	require 'user_privileges/user_privileges_' . $current_user->id . '.php';
+
```
joebordes: You should use the $user object instead of $current_user

joebordes: You have to do [something like this](https://github.com/tsolucio/corebos/blob/master/include/Webservices/UploadProductImages.php#L30:L33) here to check if the user has permission to access the module.

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,51 @@
...
+	$cur_tab_id = getTabid($module);
+	$sql1 = "select * from vtiger_relatedlists where tabid=?";
+	$result = $adb->pquery($sql1, array($cur_tab_id));
+
+	$num_row = $adb->num_rows($result);
+	for ($i = 0; $i < $num_row; $i++) {
+		$rel_tab_id = $adb->query_result($result, $i, 'related_tabid');
+		//$function_name = $adb->query_result($result, $i, 'name');
```
joebordes: You have to check here if the user has access to this module (like above), If not you should not return that array element

build/wsChanges/CPgetRelatedModules.php
```diff
@@ -0,0 +1,25 @@
...
+ * License terms of Creative Commons Attribution-NonCommercial-ShareAlike 3.0 (the License).
+ ********************************************************************************/
+$operationInfo = array(
+	'name'    => 'getRelatedModulesInfomation',
+	'include' => 'include/Webservices/getRelatedModules.php',
+	'handler' => 'getRelatedModulesInfomation',
+	'prelogin'=> 0,
+	'type'    => 'POST',
+	'parameters' => array(
+	array('name' => 'module','type' => 'string'),
```
joebordes: Please correct indentation

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,52 @@
...
+	$cur_tab_id = getTabid($module);
+	$sql1 = "select * from vtiger_relatedlists where tabid=?";
+	$result = $adb->pquery($sql1, array($cur_tab_id));
+	$num_row = $adb->num_rows($result);
+	for ($i = 0; $i < $num_row; $i++) {
+		$rel_tab_id = $adb->query_result($result, $i, 'related_tabid');
+		$label = $adb->query_result($result, $i, 'label');
+		$actions = $adb->query_result($result, $i, 'actions');
+		$relationId = $adb->query_result($result, $i, 'relation_id');
+		if (!in_array($label, $types['types'])) {
```
joebordes: You can't use the label here because there could be a related list with a label different than the module it is listing. So you will have to call getTabModuleName($rel_tab_id), BUT you can only do that if it is different than 0, so you will have to move the "if" you have below in line 40 up here and change the code a bit.

If $rel_tab_id is 0 you should just include the related list, there is no need to check for permissions in that case.

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,52 @@
...
+			continue;
+		}
+		if ($rel_tab_id != 0) {
+			if ($is_admin || $profileTabsPermission[$rel_tab_id] == 0) {
+				if ($is_admin || $profileActionPermission[$rel_tab_id][3] == 0) {
+					$focus_list[$label] = array('related_tabid' => $rel_tab_id, 'label'=> $label, 'labeli18n' =>getTranslatedString($label, getTabModuleName($rel_tab_id)), 'actions' => $actions, 'relationId' => $relationId);
```
joebordes: use the getTabModuleName($rel_tab_id) call from above in order to save the call overhead

include/Webservices/getRelatedModules.php
```diff
@@ -0,0 +1,52 @@
...
+				}
+			}
+		} else {
+			$focus_list[$label] = array('related_tabid' => $rel_tab_id, 'label'=> $label, 'labeli18n' =>getTranslatedString($label, getTabModuleName($rel_tab_id)), 'actions' => $actions, 'relationId' => $relationId);
```
joebordes: use the getTabModuleName($rel_tab_id) call from above in order to save the call overhead

joebordes:
I added some additional comments

joebordes:
Thanks!

PR #314, opened 2018-07-19
feat(Webservice) Allow WS clients to send DB formatted productline cu…

Luke1982:
…rrencies

The purpose of this PR is to allow WS clients to insert a 'special' hidden field called '__cbws_skipcurdbconv'. When set to 'true', it will force the webservice update and revise methods to **not** use currency conversion to DB format. In other words, it allows the client to send DB formatted values without having to worry about them being multiplied by a million.

### DIFF
--- a/data/CRMEntity.php
+++ b/data/CRMEntity.php
@@ -628,9 +628,11 @@ class CRMEntity {
 
 			$ajaxSave = false;
 			if ((isset($_REQUEST['file']) && $_REQUEST['file'] == 'DetailViewAjax' && isset($_REQUEST['ajxaction']) && $_REQUEST['ajxaction'] == 'DETAILVIEW'
-						&& isset($_REQUEST["fldName"]) && $_REQUEST["fldName"] != $fieldname)
-					|| (isset($_REQUEST['action']) && $_REQUEST['action'] == 'MassEditSave' && !isset($_REQUEST[$fieldname."_mass_edit_check"])
-							&& (!isset($_REQUEST['ajxaction']) || $_REQUEST['ajxaction'] != 'Workflow'))) {
+				&& isset($_REQUEST["fldName"]) && $_REQUEST["fldName"] != $fieldname)
+				|| (isset($_REQUEST['action']) && $_REQUEST['action'] == 'MassEditSave' && !isset($_REQUEST[$fieldname."_mass_edit_check"])
+				&& (!isset($_REQUEST['ajxaction']) || $_REQUEST['ajxaction'] != 'Workflow'))
+				|| (!empty($this->column_fields['__cbws_skipcurdbconv'.$fieldname]) || !empty($this->column_fields['__cbws_skipcurdbconvall'])))
+			{
 				$ajaxSave = true;
 			}
 
--- a/include/Webservices/ProductLines.php
+++ b/include/Webservices/ProductLines.php
@@ -33,7 +33,7 @@ $subtotal = 0;
 $totalwithtax = 0;
 $i = 0;
 $pdoInformation=$element['pdoInformation'];
-$skipCurDBConv = (isset($element['__cbws_skipcurdbconv']) && ($element['__cbws_skipcurdbconv'] == 'true')) ? true : false;
+$skipCurDBConv = (isset($element['__cbws_skipcurdbconv_pdo']) && (!empty($element['__cbws_skipcurdbconv_pdo'])) ? true : false;
 foreach ($pdoInformation as $pdoline) {
 	$i++;
 	$_REQUEST['deleted'.$i]=(isset($pdoline['deleted']) ? $pdoline['deleted'] : 0);
--- a/include/Webservices/ProductLines.php
+++ b/include/Webservices/ProductLines.php
@@ -33,6 +33,7 @@ $subtotal = 0;
 $totalwithtax = 0;
 $i = 0;
 $pdoInformation=$element['pdoInformation'];
+$skipCurDBConv = (isset($element['__cbws_skipcurdbconv']) && ($element['__cbws_skipcurdbconv'] == 'true')) ? true : false;
 foreach ($pdoInformation as $pdoline) {
 	$i++;
 	$_REQUEST['deleted'.$i]=(isset($pdoline['deleted']) ? $pdoline['deleted'] : 0);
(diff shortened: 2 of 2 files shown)


joebordes:
this change looks correct/safe and backward compatible but shouldn't we extend it to support ALL currency fields, not only the ones on product lines?

Luke1982:
That seems like a logical thing to do, I'll get some work done to keep it consistent

Luke1982:
I followed the path of the data. From the entry point in webservice.php, the data is passed to the runOperation method of the [OperationManager](https://github.com/tsolucio/corebos/blob/master/include/Webservices/OperationManager.php). That passes the data to either the vtws_retrieve or vtws_update function through 'call_user_func_array' on line 138. They either call revise or update in [VtigerModuleOperation](https://github.com/tsolucio/corebos/blob/master/include/Webservices/VtigerModuleOperation.php), which basically always calls the update method for both operations. In line 68 of that file, there is a 'Datatransform::sanitizeForInsert' call, that lives in [this](https://github.com/tsolucio/corebos/blob/master/include/Webservices/DataTransform.php) file. On lines [158](https://github.com/tsolucio/corebos/blob/master/include/Webservices/DataTransform.php#L158) and [159](https://github.com/tsolucio/corebos/blob/master/include/Webservices/DataTransform.php#L159) of that file, I see calls to sanitize date and currency fields have been commented out. Then back in VtigerModuleOperations, a call to update in [VtigerCRMObject](https://github.com/tsolucio/corebos/blob/master/include/Webservices/VtigerCRMObject.php#L119) is made that ultimately uses the CRMentity's save method. So as far as I can tell, there is no further Currency field conversion, but I included the path I followed so you could track it and see if I missed something. Also, if the commented lines I mentioned in the DataTansform class ever get un-commented for some reason, we **will** need to use the flag there, but maybe you know what the plan for those lines is?

Luke1982:
I see you [did this](https://github.com/tsolucio/corebos/commit/99f530f8bf4c51ab1b708b7693e485e8e8c025a8#diff-44757ed33dc9a5f4b0b0db6a9eafd53f) back in 2016 (Github is cool) to fix an error on currency formatting. But I can't make out if it was meant to either fix the case when a user input currency formatted values or when the user input database formatted values. Maybe we're better off setting a flag in the $_REQUEST global and just using it once in the CurrencyField class as a 'catch all'...

joebordes:
Your tracing is correct and the change must be made in the CRMEntity class. I created this script for [coreBOSWSDevelopment Tool](https://github.com/tsolucio/coreBOSwsDevelopment):

```
<?php
$input_array =  array (
	'amount' => '1234.56',
	'id' => '13x5744',
	'__cbws_skipcurdbconvamount' => 1
);

$dmsg.= debugmsg("Revise, sending in",$input_array);

//name of the module for which the entry has to be created.
$moduleName = 'Potentials';

$update = $cbconn->doRevise($moduleName, $input_array);

//check for whether the requested operation was successful or not.
if($update) {
	//operation was successful get the response.
	var_dump($update);
} else {
	echo "Revise failed<br>";
	$err = $cbconn->lastError();
	echo $err['code'].': '.$err['message'];
}
?>
```
That updates the amount currency field on Potentials. I changed the decimal and separator of my user and launched the script with different amount values confirming the current expected behavior. Then I tried adding the ```__cbws_skipcurdbconv``` element but I got SQL errors due to the fact that this variable affects ALL numeric and currency fields. So I tried setting all numbers to the standard US format and it worked correctly. Then I modified the code with this change to support only formatting one (or more) field.

```
diff --git a/data/CRMEntity.php b/data/CRMEntity.php
index 7d358be23..ee82ced83 100644
--- a/data/CRMEntity.php
+++ b/data/CRMEntity.php
@@ -630,7 +630,9 @@ class CRMEntity {
                        if ((isset($_REQUEST['file']) && $_REQUEST['file'] == 'DetailViewAjax' && isset($_REQUEST['ajxaction']) && $_REQUEST['ajxaction'] == 'DETAILVIEW'
                                                && isset($_REQUEST["fldName"]) && $_REQUEST["fldName"] != $fieldname)
                                        || (isset($_REQUEST['action']) && $_REQUEST['action'] == 'MassEditSave' && !isset($_REQUEST[$fieldname."_mass_edit_check"])
-                                                       && (!isset($_REQUEST['ajxaction']) || $_REQUEST['ajxaction'] != 'Workflow'))) {
+                                                       && (!isset($_REQUEST['ajxaction']) || $_REQUEST['ajxaction'] != 'Workflow'))
+                                       || (!empty($this->column_fields['__cbws_skipcurdbconv'.$fieldname]))
+                               ) {
                                $ajaxSave = true;
                        }
 
```

So, either we send in the "no convert marker" for each field or we obligate the programmer to send in ALL numeric fields for REVISE.

I understand that update will work with the generic "no convert marker" as all fields are sent in and they all must abide by the decided format.

Play around with that a little and get back to me with your toughts.

Luke1982:
I'm not sure I understand you correctly. You updated potentials and tried it with and without the marker. But in the original PR I created, I **only** affect anything in inventory modules, based on whether there are any [inventorylines](https://github.com/tsolucio/corebos/pull/314/files#diff-93171ee8dd0c96aa6a885409a177331a). That change shouldn't affect potentials, or any other module that doesn't have pdoLines, at all.

I had thought about the fact that you would be obliged to convert ALL fields to the format you selected. It seems like a good idea to enable someone to select only certain fields with the change you propose. Still this change alone would not fix the problem on inventory modules, since they are not treated as regular fields (please correct me if I'm wrong here), so this would not fix the use case I was trying to solve, which is:

You do a 'Retrieve' operation on an inventory module record and store it, let's say in a JSON object. Then you update for instance the status of that record. You then do a 'Revise' of that entire record, including the change you made to the status. The change I made would not protect any non-productline fields from this operation, but the change you made would enable someone to protect individual fields, but **not** the productlines. So basically, I think we should combine the two and rename my flag to indicate that this is only for productlines, which would be a better descriptive.

I've played around with a salesorder, but let me play around with a non-inventory module like in your example to understand what happens there and get back to you.

Luke1982:
OK, I've reproduced the example you posted and can confirm the behaviour. You will get an SQL error when you set the flag for the field but still send it in user-formatted. But personally I think that's a good thing. You're sending in data in a different form than you specify, so you should get an error.

Still we'd need **both** changes, since this will not support pdoLines. In the end I think we should maybe even have 3 options:
- ['__cbws_skipcurdbconv_pdo' : prevent converting product/service lines from undergoing currency DB conversion. I think being able to set this on a pdoLine-individual level would be a step too far. So a developer should select one format for *all* pdoLines.]
- ['__cbws_skipdbcurconv*fieldname*' : prevent converting to DB for this fieldname]
- ['__cbws_skipdbcurconvall : prevent converting to DB format for all fields, except pdoLines, which should always be stated separately]

How would you feel about that?

Luke1982:
OK, I did the following:
- I changed the first [commit](https://github.com/tsolucio/corebos/pull/314/commits/00a4058e47bdd383408ccd255c087a905377550f) to clearly indicate skipping conversion on PDO lines. This commit also follows your example by not setting the fieldvalue so restrictive.
- I added a second [commit](https://github.com/tsolucio/corebos/pull/314/commits/ae75e1ed259f7cea52e786f52879fdab99fcb0b1) that takes your base and follows that to add a flag to be able to prevent conversion on **all** fields. This **includes** things like date fields.

I tested on Potentials:
`<?php
$input_array =  array (
	'amount' => '1234.56',
	'probability' => '5.00',
	'closingdate' => '2018-07-23',
	'id' => '5x195784',
	'__cbws_skipcurdbconvall' => 1,
);

$dmsg.= debugmsg("Revise, sending in",$input_array);

//name of the module for which the entry has to be created.
$moduleName = 'Potentials';

$update = $cbconn->doRevise($moduleName, $input_array);

//check for whether the requested operation was successful or not.
if($update) {
	//operation was successful get the response.
	var_dump($update);
} else {
	echo "Update failed<br>";
	$err = $cbconn->lastError();
	echo $err['code'].': '.$err['message'];
}
?>`
I could send in both the amount and probability fields in the format I selected by setting one (or both) flags for the fields. When I used the 'all' field selector, I needed to include the closingdate in DB format to make it work, which is to be expected. The only thing I'm worried about is if the 'all' selector isn't going to conflict with an *actual* fieldname at some point.

joebordes:
Looks good to me

Thanks!

PR #315, opened 2018-07-19
fix(Autocomplete) Remove list styling on autocomplete suggestion list…

Luke1982:
… and align it with the input


joebordes:
thanks

PR #322, opened 2018-07-26
fix(Inventory) Only prevent discontinued or deleted products when cre…

Luke1982:
…ating a new record

I checked this on invoices. Set a product to 'discontinued' and copied the invoice: could not save it. Edited the existing invoice after setting the product to discontinued: could save it.


joebordes:
thanks!

PR #328, opened 2018-08-04
fix(MassEdit) Replacing empty action value with MassEditSave

tebajanga:
Replacing the empty or blank `action:""` variable with `action:"MassEditSave"`


joebordes:
Thanks!

PR #330, opened 2018-08-08
feat(Console) cronjob:create command, for creation of schedule cron jobs

albertxhani:
Hello Joe. This is another command to automate creation of cron jobs in coreBOS

screenshot - 2018-gsh-08 - 12 47 41 cest
screenshot from 2018-08-08 12 45 56


joebordes:
Nice!
Thanks for the command and the screenshots

PR #331, opened 2018-08-14
feat[coreBOS] Add Password Recovery

besartmarku:


joebordes:
We decide not to support this.

- nobody has asked for it nor supported it since it was opened over 3 years ago
- our application doesn't really need this functionality
- it is another security risk
- I would prefer to dedicate resources to implementing support for keycloak or similar

Thanks for the initiative

PR #335, opened 2018-08-21
Inventory autocomplete: fix some code formatting

Luke1982:
Joe, you mentioned [this](https://github.com/tsolucio/corebos/blob/master/include/js/Inventory.js#L1699) line as well, but what exactly is wrong with that? Just out of curiosity, do you use some kind of validation tool for this?

### DIFF
--- a/include/js/Inventory.js
+++ b/include/js/Inventory.js
@@ -1420,7 +1420,7 @@ function InventorySelectAll(mod, image_pth) {
 				}
 
 				// Build results
-				var results = this.buildResults(res);
+				this.buildResults(res);
 			}
 		},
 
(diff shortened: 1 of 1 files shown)


joebordes:
There is nothing wrong with that assignment inside the while condition, it is just to make sure that is what you intended to do instead of a '=='

Thanks!

joebordes:
I use [PHP Code Sniffer](https://github.com/squizlabs/PHP_CodeSniffer) and [ESLint](https://eslint.org/) as [explained here](http://corebos.org/documentation/doku.php?id=en:devel:developmentguidelines#code_formatting_and_validation_tools)

PR #342, opened 2018-08-22
Applying coreBOS Rule on Mass Duplication

tebajanga:
Using coreBOS Rule on Mass Dupplication.


modules/cbMap/processmap/DuplicateRelations.php
```diff
@@ -82,7 +83,10 @@ private function convertMap2Array() {
 
 		$relativemodules = array();
 		foreach ($xml->relatedmodules->relatedmodule as $r) {
-			$relativemodules[ (string)$r->module ] = (string)$r->relation;
+			$relativemodules[ (string)$r->module ] = array(
+				'relation' => (string)$r->relation,
+				'condition' => (string)$r->condition,
```
joebordes: I think this will emit a php warning if the condition is not present in the XML. Please try with a map that does not have conditions and eliminate the warning if it is there.

joebordes:
Thanks!

PR #352, opened 2018-08-31
feat(MapGenerator) Map Generator is a module used to generate Busines…

Edmondi-Kacaj:
**Create Map Generator** 

**Map Generator** is a module used to generate Business Maps automatically. The XML rows are generated after the user inserts the parameters he is going to use for a specific functionality.

In **Map Generator** you can :

1. Create and save a new map 
2. Edit an existing map 
2. Each map has a history, so you can navigate to all modification that are  made.
3. Each Business Map  has its own design.
4. The design ( GUI ) I have implemented is user friendly.

There are more than 20 map types which can be generated by this module.
There's no need to write XML rows , now you can use  **Map Generator ** to generate the Business Maps 
that we use in **coreBOS**


include/LD/assets/styles/override_lds.css
```diff
@@ -158,12 +158,6 @@ hr {
     top: -0.0625rem;
     right: -0.0625rem;
     animation-name: dotsBounceAfter-extraExtraSmall; }
-  .slds-masseditprogress {
```
joebordes: Do you have to eliminate this class? It is being used by the mass edit popup information screen. If you delete it that will probably break. Can you find another way of styling your objects that does not conflict with this class?

joebordes:
I see that you have changed the LDS CSS salesforce-lightning-design-system.css (probably updated it).

I have tried this in the past and some formatting was broken in the application. Have you validated that the whole application is working correctly after this change? Specifically, the menu, the settings, the role editor...

joebordes:
I see there are a lot of functions in All_functions.php that the application already has. We should start migrating the ones we can to the standard way of getting that information. That way we reduce the amount of code to debug and support. For the moment it is ok as it is, but please allocate some time to start asking me how we can simplify this script.

joebordes:
I see you are using Kendo in FSGUI.php, please eliminate the usage of this library as it is not open source and we cannot distribute solutions based on it.

joebordes:
I see that you have setup your own database connection. We should move that to the standard $adb API
The more code we write the harder it is to validate, debug, maintain and keep the application secure.

Start asking.

joebordes:
what is vtiger_accountinstallation?  do you create that at some point?

joebordes:
I see that you are changing some of the image files in themes/images.  Is that necessary? how does that affect the rest of the application? Why are you changing those?

include/LD/assets/styles/override_lds.css
Edmondi-Kacaj: I think this not affect anything as i know ,if you know where affect and what affect i can fix normaly.

Smarty/templates/Header.tpl
```diff
@@ -11,17 +11,21 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
 <html>
 <head>
-	<meta http-equiv="Content-Type" content="text/html; charset={$LBL_CHARSET}">
+	<meta http-equiv="Content-Type" content="text/html; charset={$APP.LBL_CHARSET}">
```
joebordes: This change is not related to your extension and it is incorrect as coreBOS does not have that string in its translation files anymore. Please eliminate this change.

Smarty/templates/Header.tpl
```diff
@@ -11,17 +11,21 @@
...
 	<meta name="robots" content="noindex">
 	<title>{$USER} - {$MODULE_NAME|@getTranslatedString:$MODULE_NAME} - {$coreBOS_app_name}</title>
 	<link REL="SHORTCUT ICON" HREF="{$FAVICON}">
 	<style type="text/css">@import url("themes/{$THEME}/style.css");</style>
 	{if $Application_JSCalendar_Load neq 0}<link rel="stylesheet" type="text/css" media="all" href="jscalendar/calendar-win2k-cold-1.css">{/if}
 	<link rel="stylesheet" href="include/print.css" type="text/css" media="print" />
 	<link rel="stylesheet" href="include/LD/assets/styles/salesforce-lightning-design-system.css" type="text/css" />
-	<link rel="stylesheet" href="include/LD/assets/styles/slds-combobox.css" type="text/css" />
+	 <link rel="stylesheet" href="include/LD/assets/styles/slds-combobox.css" type="text/css" />
+
```
joebordes: please do not add unnecessary whitespace

Smarty/templates/Header.tpl
```diff
@@ -11,17 +11,21 @@
...
 	<link rel="stylesheet" href="include/LD/assets/styles/mainmenu.css" type="text/css" />
+	<link rel="stylesheet" href="include/LD/assets/styles/LoadingAll.css" type="text/css" />
+	<link rel="stylesheet" href="include/LD/assets/styles/settings.css" type="text/css" />
 	<link rel="stylesheet" href="include/LD/assets/styles/override_lds.css" type="text/css" />
+	<link rel="stylesheet" href="include/LD/assets/styles/all-modules.css" type="text/css" />
 {* vtlib customization: Inclusion of custom javascript and css as registered *}
```
joebordes: the three changes above are related to your extension and should be loaded ONLY for that extension. First, we avoid having to load them for ALL the pages in the application when they are not needed and also we avoid unwanted side effects like some CSS rule breaking one of our existing ones.

please move these 3 to the  main template file of your extension

joebordes:
additionally to the comment above about including the three CSS files for all or only for your extension, I see that you have added all three to the base application:

```
include/LD/assets/styles/LoadingAll.css
include/LD/assets/styles/all-modules.css
include/LD/assets/styles/settings.css
```

if these files are only for your extension they should live inside that directory:

```
modules/MapGenerator/styles/LoadingAll.css
modules/MapGenerator/styles/all-modules.css
modules/MapGenerator/styles/settings.css
```

joebordes:
I see that include/LD/assets/styles/override_lds.css is still missing the slds-masseditprogress class

modules/MapGenerator/PreviewRezult.php
```diff
@@ -0,0 +1,64 @@
...
+* either express or implied. See the License for the specific language governing
+* permissions and limitations under the License. You may obtain a copy of the License
+* at <http://corebos.org/documentation/doku.php?id=en:devel:vpl11>
+*************************************************************************************************/
+
+global $adb, $log, $current_user, $default_charset;
+
+$selectedfieldsfromhistory = $_POST['queryhistory'];
+$listoffileds = explode("FROM",$selectedfieldsfromhistory);
+$myArray=explode(",",str_replace("SELECT","",$listoffileds[0]));
```
joebordes: Dangerous str_replace:
 - you don't clean the input parameter
 - you only use uppercase, so if we get the SQL in lowercase, it will not explode or str_replace correctly
 - what happens if we have a field called "unselectfield" or something similar that contains "select"?  Get it? It is much faster and safer to just cut the first 6/7 characters from the trim of the string.

modules/MapGenerator/Staticc.php
```diff
@@ -0,0 +1,19 @@
...
+
+ /**
+* this is the type of errors or debugs
+*/
+abstract  class TypeOFErrors 
+{
+	const ErrorLG="----Edmondi Kacaj--- ERROR !!!";
+	const INFOLG="----Edmondi Kacaj----- INFO !!!";
+	const WARNINGLG="---Edmondi Kacaj---- WARNING !!!";
+	const SuccesLG="----Edmondi Kacaj----- SUCCESS !!!";
```
joebordes: Those names are nice but I think it would be better to keep them standard :-)

joebordes:
please delete this directory  modules/MapGenerator/bsmSelect/examples/

joebordes:
I finished my second review. We still have unanswered this from the first round:
 - what is vtiger_accountinstallation?
 - I see also another tabe "viste" which I'm not sure what that is for. Can you explain?

joebordes:
you do not need this file: modules/MapGenerator/js/json.js 
that is a polyfill to add JSON to javascript, but all our supported browsers now have native support for that. Get rid of the file and the various "includes" you have around

modules/MapGenerator/AllRelation.php
```diff
@@ -1,42 +1,43 @@
+
 <?php
+/*
+ * @Author: Edmond Kacaj
+ * @Date: 2018-09-11 11:37:45
+ * @Last Modified by: edmondikacaj@gmail.com
+ * @Last Modified time: 2018-09-11 11:38:51
+ */
```
joebordes: This is not a License, this is you signing a file and adding some information that is totally redundant. git is already tracking who is making the change, when and the whole history MUCH better than you will ever do trying to maintain this information.

You don't have to sign your files, you have to commit a lot. If you absolutely must sign them, just put your name, all the dates and other stuff is just noise and put your name right after a well-defined license so people will know how they can use your code.

mo-ndi: Is something what Visual Code generate 
i need to check all file and to remove ??

joebordes: You can configure your IDE to not add that.
Yes, please change those.

joebordes:
Hi, I am waiting on these comments (just in case):

 - https://github.com/tsolucio/corebos/pull/352#issuecomment-419156416
 - https://github.com/tsolucio/corebos/pull/352#issuecomment-419155029

joebordes:
:+1:

joebordes:
I created a MapGenerator branch to continue development there.

joebordes:
we are going to do this another way

PR #353, opened 2018-08-31
Zero bounce

Mikel1995:


include/validation/validateEmailWithZB.php
```diff
@@ -0,0 +1,57 @@
...
+ * distributed on an "AS IS" BASIS, WITHOUT ANY WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing
+ * permissions and limitations under the License. You may obtain a copy of the License
+ * at <http://corebos.org/documentation/doku.php?id=en:devel:vpl11>
+ *************************************************************************************************/
+
+class validateEmailWithZB {
+	private static $lastErrorStatus = array();
+
+	public static function validateEmail($email) {
```
joebordes: You need to add $field as a parameter to index the field name you are validating.

include/validation/validateEmailWithZB.php
```diff
@@ -0,0 +1,57 @@
...
+		$api_key = GlobalVariable::getVariable('Zero_Bounce_API_KEY', '');
+		$emailToValidate = $email;
```
joebordes: if empty($api_key) return true;

include/validation/validateEmailWithZB.php
```diff
@@ -0,0 +1,57 @@
...
+ * either express or implied. See the License for the specific language governing
+ * permissions and limitations under the License. You may obtain a copy of the License
+ * at <http://corebos.org/documentation/doku.php?id=en:devel:vpl11>
+ *************************************************************************************************/
+
+class validateEmailWithZB {
+	private static $lastErrorStatus = array();
+
+	public static function validateEmail($email) {
+		$api_key = GlobalVariable::getVariable('Zero_Bounce_API_KEY', '');
```
joebordes: if empty($api_key) return true;

include/validation/validateEmailWithZB.php
```diff
@@ -0,0 +1,57 @@
...
+		if (isset($json['status'])) {
+			if ($json['status'] == 'valid') {
+				$valueToReturn = true;
+			} elseif ($json['status'] == 'invalid') {
+				$text = $json['sub_status'];
+			}
+		} elseif (isset($json['error'])) {
+			$text = $json['error'];
+		}
+		self::$lastErrorStatus['email'] = $text;
```
joebordes: self::$lastErrorStatus[$field] = $text;   // index the errors per field being validated

include/validation/validateEmailWithZB.php
```diff
@@ -0,0 +1,57 @@
...
+		return $valueToReturn;
+	}
+	public static function getLastErrorMsg($field) {
+		echo self::$lastErrorStatus[$field];
+	}
+}
+
+function validate_ZeroBounce($field, $fieldval, $params, $fields) {
+	$v = new validateEmailWithZB();
```
joebordes: The method is static, you do not need to instantiate, just call it:
```
return validateEmailWithZB::validateEmail($fieldd, $fieldval);
```

modules/cbMap/processmap/Validations.php
```diff
@@ -324,6 +324,12 @@ public static function formatValidationErrors($errors, $module) {
 				$error.= $err . "\n";
 			}
 		}
+		if (strpos($error, "{custommsg|") > 0) {
+			preg_match_all('/{(.*)}/', $error, $match);
+			$res = explode('|', $match[1][0]);
+			include_once 'include/validation/'.$res[1].'.php';
+			$error = call_user_func(array(__NAMESPACE__ .$res[1], $res[2]), 'email');
```
joebordes: The first "foreach" contains the fieldname that you are validating, you have to use that here, not "email"

```
foreach ($errors as $errs) {
change it to
foreach ($errors as $field => $errs) {
```

then change the hardcoded email for the variable $field

joebordes:
Thanks!

PR #364, opened 2018-09-13
fix(User) Add change set to fix error WS query on Document Folders and assign to new owner when user is deleted

klivstudiosynthesis:


joebordes:
thanks!

PR #368, opened 2018-09-18
fix(cbMap) Fill layout section for ApplicationFields block type in the Detail View Layout Mapping

klivstudiosynthesis:


joebordes:
thanks!

PR #371, opened 2018-09-21
feat(asterisk) change in the process that identifies asterisk incomin…

Mikel1995:
…g call


joebordes:
@reetp , I totally agree, this is just a quick hack to cover a business requirement we have right now.

joebordes:
Thanks!

PR #372, opened 2018-09-25
fix(Inventory) Don't hide suggestion list when AC input is not hovere…

### LINKED ISSUE #358: Product/service autocomplete only shows results when mouse is over input

Luke1982:
…d, fixes #358


joebordes:
thanks!

PR #413, opened 2018-11-29
feat(Inventory) Allow listprice in PO's to be filled with the cost pr…

Luke1982:
…ice when selecting a product

### DIFF
--- a/include/js/Inventory.js
+++ b/include/js/Inventory.js
@@ -9,7 +9,8 @@
 
 var inventoryi18n = '',
 	defaultProdQty = 1,
-	defaultSerQty = 1;
+	defaultSerQty = 1,
+	setCostPriceInPos = 0;
 
 document.addEventListener('DOMContentLoaded', function () {
 	ExecuteFunctions('getTranslatedStrings', 'i18nmodule=SalesOrder&tkeys=typetosearch_prodser').then(function (data) {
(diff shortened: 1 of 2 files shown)


joebordes:
I understand that this is to set the inventory line to cost price when using the new auto-complete functionality because selecting from the popup actually does use the cost now.  Right?

If that is the case, I would eliminate the global variable because it doesn't exist for the base functionality. I mean, now coreBOS uses the cost price on PO without asking nor implementing an option to choose via a GV, so the autocomplete should follow the same philosophy.

Another option would be to implement support for the GV everywhere, but I don't see the point as the current functionality seems to be correct.

Luke1982:
Ah, I didn't know this was already the behaviour in the popup (we've been using autocomplete for a long time). I'll eliminate the GV and just always set it to costprice in PurchaseOrders. What about multicurrency?

joebordes:
I think I fixed that a few days ago: 2fe89510c6c4dc056af117246ff1d4e06346ff58

Luke1982:
Yeah I saw that, I mean: will that also affect costprice? We don't get multiple currencies for cost price

joebordes:
I don't know

Luke1982:
Cost price doesn't even support multicurrency, so no harm there. At least: we don't use it now so we don't really break anything that isn't already broken.

Luke1982:
In my installation at least I can only select multiple currencies on the unit price field

joebordes:
yes, that is correct, unit price is the only multicurrency field

Luke1982:
So basically we could implement this I think, since we don't break any existing functionality

joebordes:
but without the GV?

Luke1982:
Yes, I took all that out and just check the gVTModule variable to see if it is a PurchaseOrder

PR #414, opened 2018-11-29
feat(i18n) Rename price column in PurchaseOrders

Luke1982:


Luke1982:
Hold up, this isn't correct...

Luke1982:
Now it is

joebordes:
looks good to me.  Thanks!

PR #416, opened 2018-12-06
feat(changeSet) add field in cbQuestion

glorymoshi:


joebordes:
manually merged because there was an error in the XML file.
I also applied some changes:
 - label translation
 - move Description block to end
 - format files
https://github.com/tsolucio/corebos/commit/c24e053d3e0fd77e83de65f46937ea04848584bf

Thanks!

PR #417, opened 2018-12-07
coreBOS Rule expression as a Validation option

tebajanga:
Adding expression on coreBOS validation rule which will take the expression and evaluate it with `coreBOS_Rule`


joebordes:
Thanks!

PR #422, opened 2018-12-09
fix(Listview) Format N and NN fields in listview the same way we do i…

Luke1982:
…n detailview

### DIFF
--- a/include/ListView/ListViewController.php
+++ b/include/ListView/ListViewController.php
@@ -403,6 +403,10 @@ class ListViewController {
 						}
 						$value = '<span style="float:right">'.$value.'</span>';
 					}
+				} elseif ($field->getFieldDataType() == 'double') {
+					if ($value != '') {	
+						$value = CurrencyField::convertToUserFormat($value);
+					}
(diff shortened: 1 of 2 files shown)


joebordes:
I understand that this is to fix the case about the list view mappings you posted in the forum

Checking only on double seems very broad. Wouldn't that affect numbers also?  Format them as currency?

Can you change the condition to check on `uitype == currency` in order to be a bit more specific?

Luke1982:
It would affect numbers and in fact, that is my goal. Right now we display numbers in detailview (anything that is UI type 7) formatted by the CurrencyField class as well. My goal here is to synchronize the behavior between detailview and listview. The checking for currency is already there, in fact I copied it and stripped out all the unwanted stuff to avoid setting totals at the bottom and other currency stuff (check the code right above the code I added). Right now this only works on listview, I'm working on doing the same in the related lists.

Luke1982:
And also, apply currency formatting to currency fields in related lists

Luke1982:
I've added the currencyfield bit to the related lists as well.

joebordes:
thanks!

PR #423, opened 2018-12-10
Change Product-Subproduct Relationship in the new module

jonasinaj:


Luke1982:
If I'm not mistaken, when you alias tables in a join you don't need the 'AS' keyword.

jonasinaj:
> If I'm not mistaken, when you alias tables in a join you don't need the 'AS' keyword.

Hello! 
Actually, both ways with 'AS' keyword and without that are correct. And in fact without 'AS' the code is easier to read, but I chose to do it with 'AS' keyword because I wanted to be precise that I have done 2 different joins for the same table and that for everyone it would be clear if they read that.

Luke1982:
Cool! Didn't know that. I can understand your thoughts there. Could you elaborate on what this commit does? I'm very interested in the new Product Relation functionality

joebordes:
This work is all about converting the internal Product-Bundle relation into a normal module. Up to now, we had a relationship between products where we could indicate the parent and the subproducts. This relation was being saved internally on the seproductsrel table. There was no possibility of adding units (it was implied that it was always one), nor could we add more information, nor report easily on the relation.

So we decided to create a new module: ProductComponent, which holds the Parent-Subproduct relation. Since it is a normal module we can now control the units and other information.

The module was created and installed a few weeks ago. This PR actually moves the relation from seproductsrel to ProductComponent

The last step we are working on is to create a price roll up controlled by a global variable. Coming soon.

PR #424, opened 2018-12-10
Update to lds

Luke1982:


Luke1982:
Consider this PR a work in progress but I'm getting there quite quickly. This PR is mainly about implementing the LDS framework without messing up our existing UI.

Luke1982:
@joebordes I think I have the bulk of it now. Could you test and see if I missed something?

joebordes:
see attached images for some comments.
in general, it looks very good, sadly I can't accept the PR because of the commit where you delete the config file, that would break all the installs. I will write that down in the wiki and send you a link soon explaining why. In the meantime, look into the fixing the images below while I do some more checking and organize a cherry-picking script.

thanks for all the hard work, I know how hard it is to do something so complex and have barely any effect at all in the application :-)   very good job.

ldsupgrade01

ldsupgrade02

ldsupgrade03

ldsupgrade04

ldsupgrade05

ldsupgrade06

ldsupgrade07

joebordes:
You can find [the explanation about config.inc.php here](http://corebos.org/documentation/doku.php?id=en:devel:contribute#configincphp_and_tabdata_controlconfig-devincphp).

Luke1982:
> see attached images for some comments.
> in general, it looks very good, sadly I can't accept the PR because of the commit where you delete the config file, that would break all the installs. I will write that down in the wiki and send you a link soon explaining why. In the meantime, look into the fixing the images below while I do some more checking and organize a cherry-picking script.
> 
> thanks for all the hard work, I know how hard it is to do something so complex and have barely any effect at all in the application :-) very good job.
> 
> ldsupgrade01
> 
> ldsupgrade02
> 
> ldsupgrade03
> 
> ldsupgrade04
> 
> ldsupgrade05
> 
> ldsupgrade06
> 
> ldsupgrade07

Thanks, it is quite tedious work but I keep the roadmap in mind as motivation. Sadly I messed up the first commit due to my obsession to be able to do a `git add .` when developing and I couldn't figure out how to ignore tabdata.php (which is why I just removed both from the cache, which was a stupid idea of me). I'll get to work on the images you sent. One thing I forgot about, could you try setting an autocomplete on some fields and see if that is all OK?

Luke1982:
@joebordes What if I use a last commit in the PR that re-instates the original config.inc.php file and tabdata.php file? Just need to figure out how to get them back into the cache. I thought removing them from the cache wouldn't remove then from the repo, but it did (I thought it would only stop tracking them).

joebordes:
I am fixing the existing classes, the message above is for your future modifications

Luke1982:
Wow, never knew the modifier separator had changed. Thanks for the heads up!

joebordes:
I manually merged and released these changes
I will dedicate some more time to it yet, but, in general, we are ready to move forward

Thanks for all your hard work

BTW, I plan to change the `cblds-*` CSS to follow LDS naming convention. That will happen very soon, ok?

Luke1982:
Sure, no problem. You mean update the double dash modifier to a single underscore?

joebordes:
I updated the override CSS. We are ready to move forward

PR #427, opened 2018-12-12
feat(CRMEntity) Create aftersave.first event that fires before workfl…

Luke1982:
…ows and other aftersaves


joebordes:
thanks!!

PR #435, opened 2019-01-03
Adding cbQuestion getChartFromAnswer method

tebajanga:
The `getChartFromAnswer` method of **cbQuestion** module, will accept the answer as the parameter, process it and returns a **chart** as a result.


joebordes:
thanks!

PR #439, opened 2019-01-17
feat(PurchaseOrder) Fill the listprice with the costprice when conver…

Luke1982:
…ting an SO to a PO.

See if you like it. I added a parameter to the `getAssociatedProducts` function, but we could also set a flag in the REQUEST superglobal if you don't want to alter the function.

### DIFF
--- a/build/changeSets/DefineGlobalVariables.php
+++ b/build/changeSets/DefineGlobalVariables.php
@@ -158,6 +158,7 @@ class DefineGlobalVariables extends cbupdaterWorker {
 				'Product_Maximum_Number_Images',
 				'Product_SubProduct_PriceRollUp',
 				'Product_SubProduct_CostRollUp',
+				'PurchaseOrder_TransferCostPrice',
 				'Workflow_Send_Email_ToCCBCC',
 				'Workflow_GeoDistance_Country_Default',
 				'Workflow_GeoDistance_ServerIP',
--- a/include/utils/EditViewUtils.php
+++ b/include/utils/EditViewUtils.php
@@ -1307,9 +1307,10 @@ function getConvertQuoteToSoObject($focus, $quote_focus, $quoteid) {
 * Param $module - module name
 * Param $focus - module object
 * Param $seid - sales entity id
+* Param $listcostprice: true will fill the list price with the cost price of the product
 * Return type is an object array
 */
-function getAssociatedProducts($module, $focus, $seid = '') {
+function getAssociatedProducts($module, $focus, $seid = '', $listcostprice = false) {
 	global $log, $adb, $currentModule;
 	$log->debug("Entering getAssociatedProducts(".$module.",".get_class($focus).",".$seid."='') method ...");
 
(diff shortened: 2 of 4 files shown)


Luke1982:
I'm not sure though how we should treat existing discounts and such. In my opinion it would be silly to transfer granted discounts from a salesorder to purchaseorder but maybe I don't see the full picture.

reetp:
I definitely think you should set a SuperGlobal for this.

We don't have prices properly set in Products as Products doesn't handle the multitude of pricing options we need.

We just copy to PO and then set our own pricing as we go.

The default action should always be 'stay the same' so you don't break expected behaviour, with the option to change if required.

Luke1982:
Not sure if I understand what you mean but the change I made **only** takes action when you convert a salesorder to a purchaseorder. Creating a purchaseorder from let's say a product or just from blank won't be affected, nor will duplicating PO's.

reetp:
> Not sure if I understand what you mean but the change I made **only** takes action when you convert a salesorder to a purchaseorder. Creating a purchaseorder from let's say a product or just from blank won't be affected, nor will duplicating PO's.

Yes I understand that - we use it a lot ourselves, but the DEFAULT action should always be leave things as they are, with an option to use the NEW option if required.

So in our case I don't want that change, and don't want to be forced to have to go and alter things. I prefer things to remain as they are. If I choose at a later date to change then I can do so at my leisure.

Imagine it is a really important action and I don't notice the update - it could make a complete mess of my system without me realising. I have seen this occur too many times where devs add in some funky new feature and force a change which then breaks stuff.

Luke1982:
Well in that case we should also create a global variable that lets you turn this on or off. The thing is, I believe the current behavior to be intrinsically wrong, since transferring a selling price from a salesorder to the buying price on a purchaseorder doesn't make sense at all to me. When would you ever purchase something at your supplier at the exact same price that you sell it for?

reetp:
Yes there should be a GV and it should default to 'Stay the same'.

It may be wrong, but that is what people are used too..... IMHO you should never change default current behaviour without a really good reason, no matter how wrong it is.

Luke1982:
I understand where you are coming from. I'm going to alter the PR to keep the current behavior as default but add a GV that lets you turn on transferring costprice when coming from Quotes, Invoices or SO's of Products/Services.

reetp:
Perfect :-) That is exactly the way to do it (at least IMHO!!!!)

Luke1982:
These last changes should do the trick: you get a GV called 'PurchaseOrder_TransferCostPrice'. By setting that to '1' you tell the application to transfer the costprice into the listprice field when going to a PurchaseOrder. I think it may be wise to also check for the mode being an empty string Joe? That way we explicitely avoid this when editing an existing record?

The GV has a built-in trick. Checking the 'In Module List' box and selecting some modules will allow you to specify exactly which source modules you want to use for this behavior. One note, it seems that when you come from products the behavior was already hardcoded. My code worked as expected there but even setting the GV to '0' made the application transfer the costprice to the PO.

Luke1982:
Also, I'm still not sure how we should handle existing discounts that came from an SO for instance.

joebordes:
This looks good to me, so I am going to merge it in. As to the discounts and per the discussion above, I'd say that we add another GV that sets them all to zero. It doesn't make much sense to copy the discounts to the PO, but we should try not to change current behaviour.

joebordes:
Thanks!

Luke1982:
I'll try and create that one as well then

PR #443, opened 2019-01-22
feat(ONDemandImportExport) to block cbupdater Import/Export when ONDe…

shibe2018:
…mand is active


joebordes:
I manually merged this one because:
- the merge conflict
- I didn't want to make changes to configod.php
- I wanted the helper function to be positive, not negated
- I prefer to use the helper function in the import/export scripts
- I wanted to make a smaller change just checking and dying at the start

Thanks

PR #447, opened 2019-01-24
feat(Calendar4You) upgrade fullcalendar library and google client lib…

loridacito:
…rary and put composer libraries in one place


joebordes:
Thanks!!

PR #454, opened 2019-02-05
feat(cbMap)add XSD file for accountname map

glorymoshi:


joebordes:
manually merged to avoid the coreBOSTests submodule.
I also added a minOccurs on the parameter to indicate that there must be at least one if the parameters directive is given

Thanks!

PR #455, opened 2019-02-07
fix(cbMap)update on XSD file for accountname map

glorymoshi:


joebordes:
manually merged
thanks!

PR #456, opened 2019-02-07
feat(cbMap)add XSD file for accountname query map

glorymoshi:


joebordes:
manually merged
thanks!

PR #460, opened 2019-02-08
fix(cbMap) update Validation XSD specification

glorymoshi:


joebordes:
Manually merged.
Thanks!

PR #465, opened 2019-02-12
fix(CRMEntity) retrieve all fieldinfo when performing updatefield task

Luke1982:


joebordes:
Thanks for this very important error catch and fix.
I had to add a[ couple more changes](https://github.com/tsolucio/corebos/commit/855a49fce836d5d3aebeff189653592655dc292e) in the same line and I did a solid set of testing but if you can do some more tests and confirm that would be nice.

Luke1982:
Yeah this one deserves a long blog post I think. This could also affect a lot of custom scripts I write that hook into event handlers I think. Never knew that retrieve_entity_info just returns blank values. It's been a few weeks since I've written the fix and I can feel the exact details slipping away already.

PR #468, opened 2019-02-13
fix(PurchaseOrder) Never overwrite price with productcost when editin…

Luke1982:
…g existing PO


joebordes:
This change seems correct to me. I reproduced the issue and then added the change, reviewed the values.
Now editing and duplicating respect the existing values which is what all other modules do and I consider correct behavior.

Thanks!

PR #477, opened 2019-02-22
fix(Workflow)Add Delete Confirmation Prompt

mocodertz:


joebordes:
Looks good. Going to give it a try.

joebordes:
I moved the strings to the workflow module and applied some code formatting, the rest worked correctly and I really like the message.

Thanks!!

PR #490, opened 2019-03-01
feat(Actions) new pattern for actions

albertxhani:
@joebordes This is a test case I was thinking of. We can use a ontroller pattern for actions( and views) instead of plain procedural php actions. 

I have created a test Action on Leads module which can be called like this : 
index.php?action=index&module=Leads&action=LeadsAjax&actionname=test&method=init

We can create multiple methods inside the action class and call each one given the method name in the 
url ( same way a controller works in a MVC). We can also do this for views and create basic view types like popups, shell views(containing header and footer by default) and so much more

Tell me what do you think


joebordes:
this looks correct and very good to me.
Besides the validate() method and the license header, what is missing to "finish" this PR?

Can you extend a little on how the "views" would look?

PR #494, opened 2019-03-06
Create the new LDS header

Luke1982:
OK so this is going to be a nice one. The auto-merge has a nice red warning. I'm putting my faith into your git skills here. Let me know if there is anything you need from me. There might be some loose ends here so let me know if you run into anything.

*EDIT*
Just checked the conflict checker, doesn't even seem to be a whole lot that is crazy bonkers.

## Some things I can tell you already
The parameter [here](https://github.com/tsolucio/corebos/pull/494/files#diff-1d02ff3b3fc7e3db4772c8149c61c8d0R5858) should be removed, it must be some remnant from when I was trying some different pattern.

Also, when the dropdown next to the global search loads on the page load, it just gives you the placeholder. Ideally you'd want that to contain the selected module name or 'Mutiple' when multiple are selected or 'None' when none are. It does that as soon as you focus on it and (de)select a module because that triggers the mechanism but it's not really clear until that point what the status of the dropdown is. It **does** respect the setting though, even when you never open the dropdown.

I dropped a bunch of CSS in style.css of which I'm not sure I should have. It seemed like old LDS-hacks from when we didn't include the full framework but I'm not 100% sure.

Ideally the 'Last viewed' panel should also move up when the page is scrolling down, but I consider that something to be done in the near future.

## User photo
The nicest thing would be if we could include the user's photo in the avatar thingy at the top right. The `$current_user` holds a reference to the file name, but not the full path though.


joebordes:
Looks great.  Thanks!!

PR #498, opened 2019-03-08
feat(Webservices) Add MassDelete webservice endpoint to delete multip…

klivstudiosynthesis:
…le records at once


joebordes:
Thanks!!

PR #499, opened 2019-03-09
fix(LDS) Update LDS to 2.8.3 and fix missing classes

Luke1982:
This adds support for form elements that I think we're going to need very much (mainly creating horizontally stacked form elements without the need to creating a grid all the time, avoiding a lot of redundant markup).

This **does** drop support for the `--` modifier in favor of the `_` one, so we might run into some trouble with the new header, not sure about that.

Apart from that I don't see many issues since most of the changes are in LDS specific classes. I've tried this in a development installation and didn't see any weird stuff but haven't tried it with the new header yet...

One other reason to update is that much of the documentation is based on the newer versions and building some layout without knowing it'll work or not is annoying.


joebordes:
thanks!

PR #501, opened 2019-03-11
fix(Validations) Adding conditions for multiple fields on notDuplicate

tebajanga:
Adding conditions to support multiple fields on notDuplicate


joebordes:
thanks!

PR #508, opened 2019-03-14
feat(TimeValidation) Allow uitype 14 fields to show and save more tha…

Luke1982:
…n 23 hours


joebordes:
following up on your comments in gitter I understand that the max value should be changed to 838?

Luke1982:
Yes, let me update the commit, forgot that

joebordes:
thanks!!

PR #524, opened 2019-03-28
Update softed style.css

Impamark:
Remove the background gif - also fixes background colour issues - add a rounded corner box to pages instead
Set small separator to left of '+' to background colour to hide it
Add rounded corners and padding to Edit View page top tab to match Detail View tabs


joebordes:
thanks!

PR #525, opened 2019-03-28
Fix no space with - youhave - in test email

Impamark:
Test email shows:
This is a test mail sent to confirm if a mail is actually being sent through the smtp server that youhave configured.

This adds a space for 'you have'


joebordes:
thanks!

PR #527, opened 2019-03-29
Additional German translations

partnerwerk:
Hi, I added more translations. Hope it helps. At least I am learning a lot :-)


joebordes:
Very good! I'm glad your having fun :-)

I merged this undoing the gitignore change because there may be installs that actually want to version that code. Actually, I would recommend doing so always.

I also removed your custom style sheet which, again, should be versioned in your private repository

Thanks!!

partnerwerk:
Thank you for fixing my mistakes Joe,

I realized I had these unwanted files in the branch when I already pushed it to the server. I deleted them in my local install and will be working on more translations from time to time. So I leave this branch for future works.

PR #533, opened 2019-04-03
SendGrid integration and dashboard

tebajanga:


joebordes:
cherry-picked
I really like the look of the settings screen
I will let you know how testing goes

thanks!

PR #537, opened 2019-04-08
feat(workflow) set many to many relation

george1806:


joebordes:
I had to make some changes. In general, it was mostly correct except for the "Del All Related" feature which was missing.

thanks!

PR #549, opened 2019-04-19
feat(Inventory) Apply the 'fillfields' directive in P/S autocomplete

Luke1982:
This update allows you to use the fillfields directive in the FieldInfo BusinessMap. For more information, check [this](https://gist.github.com/Luke1982/d886a67eb661db777d93e7e645076ecc) gist where I will update to explain [how it works](https://gist.github.com/Luke1982/d886a67eb661db777d93e7e645076ecc#third-selecting-which-fields-should-be-filled-with-which-information)


joebordes:
thanks!

PR #551, opened 2019-04-23
fix(ModuleUpdate) Make 'import_Field' public since PackageUpdate refe…

Luke1982:
…rences it

I had a problem adding fields through the manifest file doing a module update because this method needs to be called outside of its class.


joebordes:
thanks!

PR #555, opened 2019-04-24
Fieat(Home)Fine tune Aggregate Widget

glorymoshi:


joebordes:
I recovered the more link which does seem to be working and added some more translations that you had overlooked (try it another language next time)

thanks!

PR #557, opened 2019-04-25
Don't do multicurrency autocomplete in PO's, fixes #473

### LINKED ISSUE #473: Products with multiple currencies don't fill costprice in PO's
https://github.com/tsolucio/corebos/blob/b780d815f6bb00c62783011ee9dab714d1958863/include/js/Inventory.js#L1869

This lines needs another check for the gVTModule not being PurchaseOrder

Luke1982:


Luke1982:
When you would do an autocomplete search for a product in a PO and that product has multicurrency activated, the selling price would be filled. This fixes it so that in PO's, the multicurrency bit it ignored.

joebordes:
why ignore multicurrency?

Luke1982:
Because we switched to costprice entirely for PO's and that field doesn't support it. The way it is now is that the autocomplete uses costprice when MC is not active on the product but uses sellingprice when it is.

joebordes:
got it, thanks

PR #568, opened 2019-05-02
feat(cbCVManagement) add field and workflow

glorymoshi:


joebordes:
manually merged. thanks

PR #584, opened 2019-06-20
fix(cbMap) extend field info buisness map accepts more than one field definition in the same map

glorymoshi:


joebordes:
thanks!

PR #590, opened 2019-07-02
Eliminate text from labels in workflow grid

geofreybundala:


joebordes:
Next time please [follow the commit rules for the message](http://corebos.com/documentation/doku.php?id=en:devel:commitguidelines). I would have put: 

`i18n(Workflow) eliminate search text for a better i18n`

thanks!

PR #591, opened 2019-07-03
Fix reports time value error

Egiliad:


joebordes:
closing in favor of #593

PR #592, opened 2019-07-03
feat(QuickCreate) adding tax fields in both Products and Services mod…

shibe2018:
…ules

### DIFF
--- a/Smarty/templates/QuickCreate.tpl
+++ b/Smarty/templates/QuickCreate.tpl
@@ -33,10 +33,7 @@
 				</tr>
 			{/foreach}
 			{if $MODULE eq 'Products' || $MODULE eq 'Services'}
-				{if $rowiteration==2}
-					<tr><td colspan="2" class="dvtCellInfo">&nbsp;</td></tr>
-				{/if}
-				{assign var=rloopit value=$rowiteration}
+				{assign var=rloopit value=2}
 				{foreach item=tax key=count from=$TAX_DETAILS}
 					{if $rloopit==2}
 						<tr>
--- a/Smarty/templates/QuickCreate.tpl
+++ b/Smarty/templates/QuickCreate.tpl
@@ -32,6 +32,36 @@
 					{/foreach}
 				</tr>
 			{/foreach}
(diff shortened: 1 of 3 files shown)


Smarty/templates/QuickCreate.tpl
```diff
@@ -33,10 +33,7 @@
 				</tr>
 			{/foreach}
 			{if $MODULE eq 'Products' || $MODULE eq 'Services'}
-				{if $rowiteration==2}
-					<tr><td colspan="2" class="dvtCellInfo">&nbsp;</td></tr>
-				{/if}
-				{assign var=rloopit value=$rowiteration}
+				{assign var=rloopit value=2}
```
joebordes: I don't think this will work as there is no way of you knowing that it will balance correctly the row. Whether you need to close the TR or not depends on the number of fields before you get to the taxes section and that is determined by the fields activated by the admin user, so you cannot hardcode it to 2, you must know what iteration you are on.

modules/Products/QuickCreate.php
```diff
@@ -43,7 +44,43 @@
 $smarty->assign('BASE_CURRENCY', $base_currency);
 //End - add multi currency
 
-$tax_details = getAllTaxes('available');
+//Tax handling (get the available taxes only) - starts
+$smarty->assign('MODE', $focus->mode);
+if ($focus->mode == 'edit') {
```
joebordes: I think this will never happen. AFAIK QuickCreate is ALWAYS creating, you cannot edit in the popup quick create screen.

modules/Products/QuickCreate.php
```diff
@@ -43,7 +44,43 @@
...
+	$retrieve_taxes = true;
+	$productid = $focus->id;
+	$tax_details = getTaxDetailsForProduct($productid, 'available_associated');
+} elseif (isset($_REQUEST['isDuplicate']) && $_REQUEST['isDuplicate'] == 'true') {
```
joebordes: this will never happen either

joebordes:
ping.
  any update on this?

joebordes:
thanks!

PR #594, opened 2019-07-08
Field Mapping modification to support search

tebajanga:


modules/cbMap/processmap/Mapping.php
```diff
@@ -153,6 +153,21 @@ public function processMap($arguments) {
 					$ct = new VTSimpleTemplateOnData($testexpression);
 					$value.= $ct->render($entityCache, $mapping['origin'], $ofields).$delim;
 					$util->revertUser();
+				} elseif (strtoupper($idx[0])=='RULE') {
+					$mapid = array_pop($fieldinfo);
+					$fieldname = array_pop($fieldinfo);
+					if (empty($ofields['record_id'])) {
```
joebordes: shouldn't this condition be "if NOT empty"  ??

joebordes:
thanks!

PR #596, opened 2019-07-10
i18n(cbtranslation) Update getLanguage with Dutch

Luke1982:
I suppose you want to use both single and plural 'million' in something like '1 million' or '2 million'? In that case in Dutch it doesn't make a difference. In something like 'Millions of dollars' though, it does. Let me know which one is the actual use case.


joebordes:
it is correct. Thanks!!

PR #600, opened 2019-07-13
feat(Users) Adding statistical numbers to user's ListView

Egiliad:


joebordes:
this looks perfect. the only comment is that I would have assigned the status user query once to save one assignment and the memory associated to it:

``` php
$status_user ='SELECT COUNT(id) as users FROM vtiger_users WHERE status=?';
```
instead of active_user and inactive_user, and use status_user for both queries with different parameters.

Thanks!!

PR #603, opened 2019-07-15
feat(Workflow) Add cond to check if parent has this as only child

Luke1982:
OK, this works rudimentary. What this allows you to do is take a reference field in the workflow conditions and select 'has this as only child'. So if you are working on for instance assets and you want to know if this asset is the only one that the related account has, this will let you do stuff decided on if that is true.

When you select it, the input box disappears (by design) since it has no function. We could re-add it and expand the function to some sort of siblingcount? Also, I can't find how to hide the input box when entering the workflow or task, only when the picklist changes.

### DIFF
--- a/include/js/de_de.lang.js
+++ b/include/js/de_de.lang.js
@@ -419,5 +419,5 @@ var alert_arr = {
 	'ERR_Massedit' : 'Fehler bei der Massenbearbeitung',
 	'ProcessFINISHED' : 'Prozess beendet',
 	'duplicatednotallowed' : 'Duplizierte Module sind nicht erlaubt',
-	'HAS_THIS_AS_ONLY_CHILD' : 'hat diese Platte als Einzelkind'
+	'HAS_THIS_AS_NTH_CHILD' : ' Hat dieses als ein weiteres Kind'
 };
--- a/include/js/de_de.lang.js
+++ b/include/js/de_de.lang.js
@@ -418,5 +418,6 @@ var alert_arr = {
 	'OF' : 'von',
 	'ERR_Massedit' : 'Fehler bei der Massenbearbeitung',
 	'ProcessFINISHED' : 'Prozess beendet',
-	'duplicatednotallowed' : 'Duplizierte Module sind nicht erlaubt'
+	'duplicatednotallowed' : 'Duplizierte Module sind nicht erlaubt',
+	'HAS_THIS_AS_ONLY_CHILD' : 'hat diese Platte als Einzelkind'
 };
--- a/include/js/de_de.lang.js
+++ b/include/js/de_de.lang.js
@@ -419,5 +419,5 @@ var alert_arr = {
 	'ERR_Massedit' : 'Fehler bei der Massenbearbeitung',
 	'ProcessFINISHED' : 'Prozess beendet',
 	'duplicatednotallowed' : 'Duplizierte Module sind nicht erlaubt',
-	'HAS_THIS_AS_NTH_CHILD' : ' Hat dieses als ein weiteres Kind'
+	'HAS_THIS_AS_NTH_CHILD' : 'Hat dieses als ein weiteres Kind'
 };
--- a/include/js/en_gb.lang.js
+++ b/include/js/en_gb.lang.js
@@ -384,5 +384,6 @@ var alert_arr = {
 	'OF' : 'of',
 	'ERR_Massedit' : 'Error on Mass Edit',
 	'ProcessFINISHED' : 'Process Finished',
-	'duplicatednotallowed' : 'Duplicated Modules Not Allowed'
+	'duplicatednotallowed' : 'Duplicated Modules Not Allowed',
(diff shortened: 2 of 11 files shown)


joebordes:
we need to use QueryGenerator here as you cannot be sure the uitype 10 field you are searching on will be on the main table, it could be in custom field table.

[Look at the "EXISTS" code](https://github.com/tsolucio/corebos/blob/master/modules/com_vtiger_workflow/VTJsonCondition.inc#L187).

Luke1982:
True, we could use QG for that, would be a nice exercise, but would we also want a wider functionality in that we don't only check if it is an only child or not but also count the siblings? I could imagine you'd want to mark a helpdesk ticket if it was the nth on a project for example.

joebordes:
that would be very interesting and shouldn't be much more difficult than you already have as I think that at this point in the code you have the value already processed in the $value variable

Luke1982:
Let me dive into the QG, I have the code ready but not using QG

Luke1982:
OK, so I tried using the QG, but since that code isn't really documented I got a bit frustrated trying to figure it out (especially since it doesn't include the CF tables by default?). Also, I realized that you don't need to include the CF table if the field doesn't live there and in fact, you might NOT want to do that.

During development, I ran into the problem where you'd want to select only the rows where the parent ID was present on the column of one of the tables (either the regular table or the CF table). But since you don't know which table you need, you have to create an OR clause which will fail because you're always going to check on a column that doesn't exist (the column lives on one table and one table alone, but you have to check both since you don't know which one).

So I changed it to use stock vtiger code to get the correct table for that field.

Now, you can select the new condition (is the nth child of) when selecting a reference field, set the type to 'formula' and then just enter the number. The code then checks if this child's parents has the N amount of children like this record and if this child is the nth child (checked against the number you enter) it will return true, otherwise false.

Basically, it will allow you to take action when any record is the Nth child of the parent, based on the parent field you select.

joebordes:
query generator has a [whole bunch of examples in the unit tests](https://github.com/tsolucio/coreBOSTests/blob/master/include/QueryGenerator/QueryGeneratorTest.php).
QG is rather intelligent, you don't have to worry about where the field is, QG will take care of that for you. You give it the module, the fields you want to see and the conditions, it will take care of the rest.
The code you need is something like this:

``` PHP
$queryGenerator = new QueryGenerator($entityData->getModuleName(), $current_user);
$queryGenerator->setFields(array('id'));
$queryGenerator->addCondition($cond['fieldname'], $parent_id, 'e');
$query = $queryGenerator->getQuery().' ORDER BY vtiger_crmentity.createdtime ASC';
```

that's it

Do you want to give that a try or should I just accept the PR, as it looks correct and the logic seems right?

Luke1982:
Let me give that a whirl, just a sec

Luke1982:
Ah, now I remember. QueryGenerator gave me
```sql
SELECT vtiger_servicejob.servicejobid FROM vtiger_servicejob  INNER JOIN vtiger_crmentity ON vtiger_servicejob.servicejobid = vtiger_crmentity.crmid LEFT JOIN vtiger_assets  ON vtiger_servicejob.related_asset_id = vtiger_assets.assetsid  WHERE vtiger_crmentity.deleted=0 AND ( trim(vtiger_assets.assetname) = '202318')  AND vtiger_servicejob.servicejobid > 0 ORDER BY vtiger_crmentity.createdtime ASC
```
Where it would try to find the assetname (my testparent is assets) from the assets table in stead of looking for the related_asset_id in a table of a module I wrote (vtiger_servicejob). I couldn't figure out how to fix that.

joebordes:
IDs are special as they try to be intelligent to make list view searches easier.
try changing the condition to the related module condition:

```
$queryGenerator->addReferenceModuleFieldCondition($child_setype, $cond['fieldname'], 'id', $parent_id, 'e');
```

joebordes:
https://github.com/tsolucio/coreBOSTests/blob/master/include/QueryGenerator/QueryGeneratorTest.php#L312

Luke1982:
Hmm, now it still tries to find the child ID, not the parent ID
```sql
SELECT vtiger_servicejob.servicejobid FROM vtiger_servicejob INNER JOIN vtiger_crmentity ON vtiger_servicejob.servicejobid = vtiger_crmentity.crmid LEFT JOIN vtiger_servicejob AS vtiger_servicejobrelated_asset_id ON vtiger_servicejobrelated_asset_id.servicejobid=vtiger_servicejob.related_asset_id WHERE vtiger_crmentity.deleted=0 AND (vtiger_servicejobrelated_asset_id.servicejobid = '202318') AND vtiger_servicejob.servicejobid > 0 ORDER BY vtiger_crmentity.createdtime ASC 
```

joebordes:
look at the parent/child on the relation I am just guessing, it may be the reverse

Luke1982:
Let me toy with it later when I have some more time. Will get back on this.

Luke1982:
OK let's just accept it as it is. I don't see myself having an extra hour and I need this functionality...

joebordes:
done!
thanks!

PR #604, opened 2019-07-16
Convert eliminate uitype

shibe2018:


build/changeSets/2019/convertUitype20to19.php
```diff
@@ -0,0 +1,34 @@
...
+	public function applyChange() {
+		global $adb;
+		if ($this->hasError()) {
+			$this->sendError();
+		}
+		if ($this->isApplied()) {
+			$this->sendMsg('Changeset '.get_class($this).' already applied!');
+		} else {
+            $tabid = getTabid('Faq');
+            $this->ExecuteQuery("UPDATE vtiger_field SET uitype = 19 WHERE tablename = 'vtiger_faq' and tabid =? and (columnname = ? or columnname = ?)", array($tabid, 'question', 'answer'));
```
joebordes: don't we have to eliminate the uitype from webservice?
wouldn't it be better to not be so specific? I mean, imagine a coreBOS install where someone has created a new field of uitype 20, with your change that field would not be updated and would stop working. I think it would be better to just change all the 20 to 19 regardless of what module they are on.  Do you agree?

build/changeSets/2019/convertUitype22and24to21.php
```diff
@@ -0,0 +1,42 @@
...
+            $tabid_qoutes = getTabid('Quotes');
+            $tabid_purchaseorder = getTabid('PurchaseOrder');
+            $tabid_salesorder = getTabid('SalesOrder');
+            $tabid_invoice = getTabid('Invoice');
+            $this->ExecuteQuery("UPDATE vtiger_field SET uitype = 21 WHERE tabid =? and (columnname = ?)", array($tabid_helpdesk, 'title'));
+            $this->ExecuteQuery("UPDATE vtiger_field SET uitype = 21 WHERE tabid =? and (columnname = ? OR columnname = ?)", array($tabid_qoutes, 'bill_street', 'ship_street'));
+            $this->ExecuteQuery("UPDATE vtiger_field SET uitype = 21 WHERE tabid =? and (columnname = ? OR columnname = ?)", array($tabid_purchaseorder, 'bill_street', 'ship_street'));
+            $this->ExecuteQuery("UPDATE vtiger_field SET uitype = 21 WHERE tabid =? and (columnname = ? OR columnname = ?)", array($tabid_salesorder, 'bill_street', 'ship_street'));
+            $this->ExecuteQuery("UPDATE vtiger_field SET uitype = 21 WHERE tabid =? and (columnname = ? OR columnname = ?)", array($tabid_invoice, 'bill_street', 'ship_street'));
+			$this->sendMsg('Changeset '.get_class($this).' applied!');
```
joebordes: same comment as the previous changeset

build/changeSets/2019/convertUitype20to19.php
shibe2018: yes I agree that mean this was wrong because it only look at module FAQ mean specific to only module

build/changeSets/2019/convertUitype22and24to21.php
shibe2018: yes that too I see

build/changeSets/2019/convertUitype20to19.php
```diff
@@ -0,0 +1,33 @@
...
+	public function applyChange() {
+		global $adb;
+		if ($this->hasError()) {
+			$this->sendError();
+		}
+		if ($this->isApplied()) {
+			$this->sendMsg('Changeset '.get_class($this).' already applied!');
+		} else {
+			$this->ExecuteQuery("UPDATE vtiger_field SET uitype =19 WHERE uitype =?", array(20));
+			$this->sendMsg('Changeset '.get_class($this).' applied!');
```
joebordes: you are missing the web service field and reference type elimination, [something like this](https://github.com/tsolucio/corebos/blob/master/build/changeSets/2018/migrateUIType59to10.php#L34:L40)

build/changeSets/2019/convertUitype22and24to21.php
joebordes: same comment as the previous changeset about web service

PR #606, opened 2019-07-16
fix(Webservices) coreBOS Rule web service call missing context

Egiliad:


joebordes:
something is missing here.

[this is the commit](https://github.com/tsolucio/coreBOSTests/blob/master/include/QueryGenerator/QueryGeneratorTest.php) where we added this web service method, you will see that [there is a file that registers the method](https://github.com/tsolucio/corebos/blob/master/build/wsChanges/cbRule.php) in the database. This file defines only one parameter, so you have to modify that file for the new installs AND create a changeset to update all the installs that already have it defined.

joebordes:
thanks!

Issue #614, opened 2019-08-05
Global search gets autocomplete target when no AC is set


Luke1982:
When no autocomplete is set for the global search but one is set on a field, the global search still has an element with the class name 'relation-autocomplete__target'. When you perform a search on the field you set the autocomplete on, the search results will show up on the global search input, not on the input you set the autocomplete on.

PR #615, opened 2019-08-05
fix(Autocomplete) Only render resultbox when global AC is active, fix…

Luke1982:
…es #614

### DIFF
--- a/Smarty/templates/Header.tpl
+++ b/Smarty/templates/Header.tpl
@@ -149,9 +149,11 @@
 												</svg>
 											</span>
 										<input name="query_string" id="query_string" class="slds-input slds-combobox__input{if $GLOBAL_AC} autocomplete-input{/if}" type="text" role="textbox" placeholder="{$APP.LBL_SEARCH_TITLE}{$coreBOS_app_name}" aria-autocomplete="list" autoComplete="off" data-autocomp='{$GS_AUTOCOMP|@json_encode}' />
+										{if $GLOBAL_AC}
 										<div role="listbox" class="">
 											<ul class="slds-listbox slds-listbox_vertical slds-dropdown slds-dropdown_fluid relation-autocomplete__target" style="opacity: 0;display:block;visibility: visible;" role="presentation"></ul>
 										</div>
+										{/if}
 									</div>
 								</div>
 							</form>


joebordes:
two questions come to mind with this change:

1.- [on this line](https://github.com/tsolucio/corebos/blob/master/Smarty/templates/Header.tpl#L133) we also check for a global variable, shouldn't your "if" do the same?

2.- with this change you are only hiding the global search div if it isn't active, what happens if we have both active. I would expect the normal field to have its own dropdown and the global search have it's own, but if we don't change the CSS class it won't behave correctly in that case, no?

Luke1982:
The classes `autocomplete-input` and `relation-autocomplete__target` **only** pertain to the autocomplete functionality. The `slds-listbox` I hide in this change (when global autocomplete is not active) is purely the container for the search results for the autocomplete, so when the global autocomplete is not active, it has no further function and in fact, disrupts the functionality of other autocomplete fields on the page. So really adding `|| $Application_Global_Search_Active` to the 'if' would re-introduce the error I fixed.

About no. 2, I'm not sure what you mean. The autocomplete looks for all the inputs with a class of `autocomplete-input` and create a new autocomplete JS object for each of them, where the `ul` with a class of `relation-autocomplete__target` with the same index on the page gets set as the 'searchresult container'. I'm not hiding the entire global search div, just all the stuff that pertains to global autocomplete when there is no global autocomplete set.

joebordes:
my question is: what happens when $GLOBAL_AC is true AND you have an autocomplete set on a field?
We have to make sure it is working in all 3 cases:
 - AC on search and no field
 - no AC on search and AC on field
 - AC on both search and field

Luke1982:
The autocomplete part of the DOM needs to travel in pairs: one [input](https://github.com/tsolucio/corebos/blob/master/include/js/general.js#L4996) and one [target](https://github.com/tsolucio/corebos/blob/master/include/js/general.js#L5014). The JS code then matches them up purely based on occurance in the DOM, so it assumes the Nth target is paired with the Nth target.

The code we had rendered a target for the global AC, but not the input. That made the input on a field pair up with the target on the global AC. As far as I can see the fields do the same: create a pair of [input](https://github.com/tsolucio/corebos/blob/master/Smarty/templates/EditViewUI.tpl#L123) and [target](https://github.com/tsolucio/corebos/blob/master/Smarty/templates/EditViewUI.tpl#L141) only when the AC is turned on for that field (the same for [UI2](https://github.com/tsolucio/corebos/blob/master/Smarty/templates/EditViewUI.tpl#L153-L172).

So basically, this makes sure they always travel in pairs, which, to answer you question should result in correct behavior in every situation.

joebordes:
thanks!

PR #616, opened 2019-08-07
fix(CustomView) publicfilter to allow user to approve filter after edit it

glorymoshi:


joebordes:
manually merged to avoid the extra space, empty file and extra commit

thanks

PR #618, opened 2019-08-10
Update ListViewController.php

reetp:
A couple of minor tweaks to give the Totals column a little bit more space separation on the right - the numbers wind up being very close to the adjacent column and this just gives a bit more readability (IMHO)

The main number was already at 6px so that goes up to 10px and then line up the main Total at the bottom of the page.

### DIFF
--- a/include/ListView/ListViewController.php
+++ b/include/ListView/ListViewController.php
@@ -390,7 +390,7 @@ class ListViewController {
 						} else {
 							$value = CurrencyField::convertToUserFormat($value);
 						}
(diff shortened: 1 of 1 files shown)


joebordes:
manually merged to avoid missing closing quote and include the same change on the related list view

eead8e1d9ab21c3d006be3e84363a83636def55e

thanks!

reetp:
Damn. Sorry - it was late :-(

joebordes:
I know that one all too well :-)

it is perfect.

PR #619, opened 2019-08-12
fix(Installation)add autocreate config.inc.php in case  doesnt exist

geofreybundala:


joebordes:
Next time please separate the formatting from the change. Always make pure formatting changes in one commit and prefix the comment with "cbSR", that way we know that we can ignore the change when we are searching for errors.
thanks!

geofreybundala:
noted.

PR #622, opened 2019-08-23
Load msg template

shibe2018:


joebordes:
cherry-picked 693e1548dec364a98d3fc530e85d94c9a7edb364 which is the only commit we needed

thanks!

PR #632, opened 2019-08-30
fix(Users) Allow read global variable before login

gaardam:
so that AD/LDAP Authentication will work

As per discussion at https://discussions.corebos.org/showthread.php?tid=1558

Is there anything else needed here?


joebordes:
Looks perfect to me.  Thanks!!

PR #635, opened 2019-09-04
Show Users List Pagination

mocodertz:


mocodertz:
The last Commit is all we need

joebordes:
manually merged to avoid master merge commits. Next time, before creating a new branch execute "git pull"

PR #639, opened 2019-09-12
fix(Issues_OnEdit_Report) fixing on edit report data disappeared

shibe2018:


joebordes:
Thanks @shibe2018 
Next time, the commit message should be:

what(where) how
in this case
fix(Reports) ....

shibe2018:
Thanks I understand

PR #641, opened 2019-09-13
fix(Autocomplete) Try converting to number before checking type, minc…

Luke1982:
…harstosearch was not working


joebordes:
thanks!!

PR #665, opened 2019-10-01
fix(Calendar) Deactivate activityreminder debug global variable

denaldd:


joebordes:
Hi

the changeset is not needed. since we are constantly adding global variables we have a Continuos changeset where we add them all. This avoids cluttering the changests directory more than it already is and also documents them all together.

So you just have to [add your new GV here](https://github.com/tsolucio/corebos/blob/master/build/changeSets/DefineGlobalVariables.php#L37)

I will make a few inline comments also on the change itself.

Finally, you have to [add the definition to the help files](https://github.com/tsolucio/corebos/blob/master/modules/GlobalVariable/language/en_us.gvdefs.php#L93)

include/js/general.js
```diff
@@ -3157,7 +3157,13 @@ function ActivityReminderRemovePopupDOM(id) {
 
 /* ActivityReminder Customization: Pool Callback */
 var ActivityReminder_regcallback_timer;
-var ActivityReminder_Deactivated = false;
+var ActivityReminder_Deactivated;
```
joebordes: leave the assignment to "false" because the call to get the correct value is ajax and takes some time so we don't want to have the code executing without a value before the variable gets assigned.

joebordes:
thanks !!

PR #668, opened 2019-10-03
Integration of function "getWeekendDates" between two dates.

roidev:
Starting from two dates "dateStart" and "dateEnd", calculate all the dates that fall on the weekend.


joebordes:
thanks!

PR #683, opened 2019-10-28
REST/SOAP fix warnings and UI for IOMap

denaldd:


denaldd:
IOMap shouldn't, so I have remove it from the UI.
I have verified also the WebserviceMapping class, but the only method I find is processMap, which requires only 1 argument.
I also made some changes to the RunWebserviceWorkflowTask class.

joebordes:
processMap argument is an array, if you look at the references for that array here

https://github.com/tsolucio/corebos/blob/master/modules/cbMap/processmap/WebserviceMapping.php#L196

you will see that the code expects [0] and [1], two parameters inside the $argument array, but only [one value is given](https://github.com/tsolucio/corebos/blob/master/modules/com_vtiger_workflow/tasks/RunWebserviceWorkflowTask.inc#L47)

joebordes:
thanks!
get back to you soon

PR #686, opened 2019-11-09
Multiple conditions gendoc loop

Luke1982:
OK so this is a pretty early thing, but it seems to work. Mainly posted it here so you could evaluate the direction.

### DIFF
--- a/include/utils/CommonUtils.php
+++ b/include/utils/CommonUtils.php
@@ -2707,18 +2707,13 @@ function getUItypeByFieldName($module, $fieldname) {
 function getTypeOfDataByFieldName($module, $fieldname) {
 	global $log, $adb;
 	$log->debug('> getTypeOfDataByFieldName ' . $module);
-	$tabIdList = array();
-	//To find tabid for this module
-	$tabIdList[] = getTabid($module);
-	if ($module == 'Calendar') {
-		$tabIdList[] = getTabid('Events');
-	}
-	$sql = 'select typeofdata from vtiger_field where tabid IN (' . generateQuestionMarks($tabIdList) . ') and fieldname=?';
-	$result = $adb->pquery($sql, array($tabIdList, $fieldname));
+	$sql = 'select typeofdata from vtiger_field where tabid = ? and fieldname=?';
+	$result = $adb->pquery($sql, array(getTabid($module), $fieldname));
 	$tod = explode('~', $adb->query_result($result, 0, 'typeofdata'))[0];
 	$log->debug('< getTypeOfDataByFieldName');
 	return $tod;
 }
+
 
 /**
  * Function to get the UItype for a field.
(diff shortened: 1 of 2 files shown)


joebordes:
this looks very interesting and well done.
I am going to add some comments to see if we can get this merged into the main branch
Can you please paste here a few examples of the syntax?

include/utils/CommonUtils.php
```diff
@@ -2699,6 +2699,27 @@ function getUItypeByFieldName($module, $fieldname) {
...
+ * Takes the input as $module - module name,and fieldname of the field
+ * returns the type of data, varchar type
+ */
+function getTypeOfDataByFieldName($module, $fieldname) {
+	global $log, $adb;
+	$log->debug('> getTypeOfDataByFieldName ' . $module);
+	$tabIdList = array();
+	//To find tabid for this module
+	$tabIdList[] = getTabid($module);
+	if ($module == 'Calendar') {
```
joebordes: We do not need this anymore. Calendar does not exist anymore and if it ever does again it will be a normal module. You can use a unique value for tabid instead of an array which will be faster.

include/utils/CommonUtils.php
```diff
@@ -2699,6 +2699,27 @@ function getUItypeByFieldName($module, $fieldname) {
...
+		$tabIdList[] = getTabid('Events');
+	}
+	$sql = 'select typeofdata from vtiger_field where tabid IN (' . generateQuestionMarks($tabIdList) . ') and fieldname=?';
+	$result = $adb->pquery($sql, array($tabIdList, $fieldname));
+	$tod = explode('~', $adb->query_result($result, 0, 'typeofdata'))[0];
```
joebordes: I would like you to save the result of the explode in an array and access the [0] element from there as we still have some clients on 5.6 and I don't remember when this syntax was introduced

modules/evvtgendoc/compile.php
```diff
@@ -678,11 +744,18 @@ function eval_paracada($condition, $id, $module, $check = false) {
...
+		$json_condition = make_json_condition($cond_elements[1], $cond_elements[2]);
+		$comp = 'wfeval';
+	}
+
 	for ($i=0; $i<count($condition_pair); $i++) {
 		$condition_pair[$i] = trim($condition_pair[$i]);
 	}
 
 	$token_pair = explode('.', $condition_pair[0]);
+	$token_pair[0] = explode(' ', $token_pair[0])[0];
```
joebordes: this looks dangerous, I understand that you are separating the conditions and it will do nothing if they are not there but I would prefer that you put this inside the "Multiple Conditions" check you do above, just to be safe.
The same comment about array access that I indicated above applies here too.

joebordes:
Note to self: migrate getDataType to getTypeOfDataByFieldName

Luke1982:
> this looks very interesting and well done.
> I am going to add some comments to see if we can get this merged into the main branch
> Can you please paste here a few examples of the syntax?

`{foreach CobroPago (paymentcategory=MyPaymentCat && credit=1)}`
Would be a syntax that works to get only the CobroPago records of that category and are marked credit. Most rudimentary examples work, we'll run into edge cases I presume at some point but I think those should be challenged when met.

modules/evvtgendoc/compile.php
Luke1982: It's been fixed. If I remember correctly the problem here was that when you use multiple conditions, a space will remain in the module name. Normally the modulename is immediately followed by a period/dot, but in the multiple conditions syntax there will most likely be an unwanted space that will prevent code further down to use the module name

Issue #688, opened 2019-11-12
Listview columnsearch is funky when column holds UI10 fields


Luke1982:
This is one I seem to be the only one having trouble with, but I'd really like to try and get work done on this soon. Since autocomplete was introduced on the searchboxes at the top of the columns in listview, it all goes to hell when that column holds UI10 fields. Need to do more testing to see what parameters you need to reproduce but I'll dig into this soon.

Luke1982:
![Peek 2019-11-12 21-49](https://user-images.githubusercontent.com/8875222/68709660-75ee3480-0596-11ea-9ac0-e31a560ceb65.gif)
Here's a peek that shows that listview column searching doesn't work at all for the vendors column in PO's. This is a clean installation with the test DB.

joebordes:
could it be related to the filter? => does it work with the ALL filter?
what is the SQL being generated?
any error in the debug log file?

Luke1982:
> could it be related to the filter? => does it work with the ALL filter?

The peek gif is a stock installation with the test DB, using the ALL  filter.

> what is the SQL being generated?
> any error in the debug log file?

Let me check those and report back. I have another installation where this happens. Manually reverting [this](https://github.com/tsolucio/corebos/commit/27a5c2c38ec790c8bd3f57edcbe9502f5adc6c4d#diff-802db39a537b3facadf682961e6038fa) commit fixed it so that's where I'm going to start searching.

Luke1982:
It seems the error occurs because there is some SQL that is triggered **before** the actual search SQL is triggered. For instance:
```sql
[...]

PR #689, opened 2019-11-12
fix(Listviewcolumnsearch) hardcode mincharstosearch to even make the …

Luke1982:
…autocomplete work at all, related to #688

This PR is live and needs updates

### DIFF
--- a/Smarty/templates/ListViewSearchBlock.tpl
+++ b/Smarty/templates/ListViewSearchBlock.tpl
@@ -186,7 +186,7 @@
 					<tr>
 						<td class="small">
 							<select name="op_cond_{$val}" id="op_cond_{$val}" class="repBox" style="width:100px;" onblur="document.getElementById('div_{$arr.fieldname}').style.display='none'">
-								{if $arr.fieldtype eq 'text'}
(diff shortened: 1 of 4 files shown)


joebordes:
couldn't we get the current value of the global variable and put it there instead of "1"?  wouldn't that solve the problem?

Luke1982:
Yes it would, forgot we had that. Let me alter the PR

Luke1982:
Just checking the list of available GV's but I can't find one that specifically controls the min chars to search

joebordes:
sorry, my bad, I am mixing the Global Search autocomplete with the list view one. You are right there is no global variable and the autocomplete map has it hardcoded too. So, besides adding the GV which, if you can't it is ok, at least, set the default value to the same default value [we have in the autocomplete map (3)](https://github.com/tsolucio/corebos/blob/master/modules/cbMap/processmap/GlobalSearchAutocomplete.php#L46) and I will accept the PR

Luke1982:
I think I'll add the GV. Since this PR is related to issue #688 I've added some more fixes for columnsearch in reference fields. I also want to style the results list so that it's as wide as the input it fills. Then I think I've solved all the funkyness

joebordes:
thanks!

PR #691, opened 2019-11-19
Executing external script on cbQuestion

tebajanga:


modules/cbQuestion/actions/qactions.php
```diff
@@ -301,5 +301,24 @@ public function delMViewWF() {
 		$rdo['notify'] = $smarty->fetch('applicationmessage.tpl');
 		echo json_encode($rdo);
 	}
+
+	public function executeScript() {
+		$record = $this->checkQIDParam();
+		$dbUser = isset($_REQUEST['db_user']) ? vtlib_purify($_REQUEST['db_user']) : '';
+		$dbPass = isset($_REQUEST['db_pass']) ? vtlib_purify($_REQUEST['db_pass']) : '';
+		$dbName = isset($_REQUEST['db_name']) ? vtlib_purify($_REQUEST['db_name']) : '';
```
joebordes: these three variables should be read from config.inc.php, so change it here and eliminate them from the JS call

modules/cbQuestion/actions/qactions.php
```diff
@@ -301,5 +301,24 @@ public function delMViewWF() {
...
+		$tablename = isset($_REQUEST['tablename']) ? vtlib_purify($_REQUEST['tablename']) : '';
+		$scriptPath = isset($_REQUEST['script_path']) ? vtlib_purify($_REQUEST['script_path']) : '';
+
+		$command = sprintf("".$scriptPath." %s %s %s %s %s", $dbUser, $dbPass, $dbName, $tablename, $record);
+		$output = shell_exec($command);
```
joebordes: we have to add a list of authorized scripts here as this is to dangerous to leave open like this. I propose a simple json/xml or a database table with the list of authorized scripts

modules/cbQuestion/actions/qactions.php
```diff
@@ -301,5 +301,31 @@ public function delMViewWF() {
 		$rdo['notify'] = $smarty->fetch('applicationmessage.tpl');
 		echo json_encode($rdo);
 	}
+
+	public function executeScript() {
+		$record = $this->checkQIDParam();
+		$db_username = $dbconfig['db_username'];
```
joebordes: don't you have to add a "global $dbconfig" here for that array to be available?

tebajanga: @joebordes , I included the global variables.

joebordes:
thanks!!

PR #699, opened 2019-12-29
feat(BussinessActions) adding parameter in link url

shibe2018:


modules/BusinessActions/BusinessActions.php
```diff
@@ -299,9 +299,25 @@ public static function getAllByType($tabid, $type = false, $parameters = false,
 			}
 
 			if ($parameters) {
-				$link->linkurl = $strtemplate->merge($link->linkurl);
-				$link->linkicon= $strtemplate->merge($link->linkicon);
-			}
+                $link->linkurl = $strtemplate->merge($link->linkurl);
+                $link->linkicon= $strtemplate->merge($link->linkicon);
```
joebordes: Don't take this out of the "if" as it will be slower if there are no parameters

modules/BusinessActions/BusinessActions.php
```diff
@@ -299,9 +299,25 @@ public static function getAllByType($tabid, $type = false, $parameters = false,
...
+					$focus_bs = CRMEntity::getInstance($parameters['MODULE']);
+					$focus_bs->retrieve_entity_info($parameters['RECORD'], $parameters['MODULE']);
+					$data_retrieve = $focus_bs->column_fields;
+					$str = $link->linkurl;
+					$subtring_start = strpos($str, '$'); 
+					$subtring_start += strlen('$');   
+					$size = strpos($str, '$', $subtring_start) - $subtring_start;   
+					$substring = substr($str, $subtring_start, $size);
+					$match = str_replace('$'.$substring.'$', $data_retrieve[$substring], $link->linkurl);
+					include_once 'include/Webservices/Create.php';
```
joebordes: Eliminate all this block, it is not necessary, VTSimpleTemplate does that and more.
Instead of including Create.php include directly what you need:
`include_one 'modules/com_vtiger_workflow/VTEntityCache.inc';`

modules/BusinessActions/BusinessActions.php
```diff
@@ -299,9 +299,25 @@ public static function getAllByType($tabid, $type = false, $parameters = false,
...
+					$entityCache = new VTEntityCache($current_user);
+					$ct = new VTSimpleTemplate($match, false);
```
joebordes: change $match to $link->linkurl

modules/BusinessActions/BusinessActions.php
```diff
@@ -299,9 +299,25 @@ public static function getAllByType($tabid, $type = false, $parameters = false,
...
+					$link->linkurl = $ct->render($entityCache, vtws_getEntityId($parameters['MODULE']).'x'.$parameters['RECORD']);
+				}
+            }
 			if ($multitype) {
```
joebordes: Correctly format the code
Note that VTSimpleTemplate uses the workflow syntax, not the limited format of the StringTemplate class, so you will have to test with the workflow field syntax.
Please create a new PR

modules/BusinessActions/BusinessActions.php
shibe2018: ok am on it

modules/BusinessActions/BusinessActions.php
```diff
@@ -299,8 +299,23 @@ public static function getAllByType($tabid, $type = false, $parameters = false,
 			}
 
 			if ($parameters) {
-				$link->linkurl = $strtemplate->merge($link->linkurl);
-				$link->linkicon= $strtemplate->merge($link->linkicon);
+				if (!empty($parameters['RECORD'])) {
+					$link->linkurl = $strtemplate->merge($link->linkurl);
+					$link->linkicon= $strtemplate->merge($link->linkicon);
```
joebordes: these two lines have to be outside this "if" because we will have situations where we have only the module to change

modules/BusinessActions/BusinessActions.php
```diff
@@ -299,8 +299,23 @@ public static function getAllByType($tabid, $type = false, $parameters = false,
...
+					$focus_bs = CRMEntity::getInstance($parameters['MODULE']);
+					$focus_bs->retrieve_entity_info($parameters['RECORD'], $parameters['MODULE']);
+					$data_retrieve = $focus_bs->column_fields;
+					$str = $link->linkurl;
+					$subtring_start = strpos($str, '$');
+					$subtring_start += strlen('$');
+					$size = strpos($str, '$', $subtring_start) - $subtring_start;
+					$substring = substr($str, $subtring_start, $size);
+					$link->linkurl = str_replace('$'.$substring.'$', $data_retrieve[$substring], $link->linkurl);
```
joebordes: delete all this code

shibe2018: YES it work correctly

shibe2018:
how that?

PR #704, opened 2020-01-08
webservice workflow task: incomplete update process

denaldd:


joebordes:
looks pretty good, a step in the right direction, but you need [to read this article](http://blog.corebos.org/blog/feedingdatatosave) and prepare the data coming from retrieve_entity_info before saving it.

Do that and let me know so I can check it again.
Thanks!

denaldd:
Please check it now.

joebordes:
that looks correct
Thanks!!

PR #705, opened 2020-01-10
feat(PageHeader) Convert pageheader to sticky LDS version

Luke1982:
One issue already is that when you add an announcement in chrome, the entire lower part of the pageheader is empty.


Luke1982:
Also, during editing, the 'edit' and 'save' buttons can also be moved into the page header. At least the top two

Luke1982:
Still have an issue when inline-editing. The header height is not taken into account.

Luke1982:
TODO need to do the 'this record had been removed' message

Luke1982:
TODO, take preheader into account https://github.com/tsolucio/corebos/blob/master/Smarty/templates/Header.tpl#L290-L292

joebordes:
THANK YOU !!!

Incredible :-D

PR #710, opened 2020-01-17
feat(Gendoc) Allow sorting in foreach loop

Luke1982:
This should allow you to do any of the following:
`{foreach Timecontrol [title=MyTitle && tcunits>5] *date_start DESC*}`
`{foreach Timecontrol *date_start DESC* [title=MyTitle && tcunits>5]}`
`{foreach Timecontrol *date_start DESC*}`

Effectively allow you to add sorting by doing `*FIELDNAME SORTORDER*` anywhere in the foreach statement and it should work.

### DIFF
--- a/include/utils/CommonUtils.php
+++ b/include/utils/CommonUtils.php
@@ -2732,6 +2732,20 @@ function makeRandomPassword() {
 }
 
 /**
+ * Function to get the columnname for a certain fieldname, given
+ * the fieldname and the module name
+ */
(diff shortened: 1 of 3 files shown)


joebordes:
thanks for bearing with me :-)

Luke1982:
> 
> 
> thanks for bearing with me :-)

:smile: no problem. Better to overdiscuss sometimes and get it right.

PR #712, opened 2020-01-21
fix(Gendoc) Also userformat NN fields, fixes #711

### LINKED ISSUE #711: GenDoc doesn't format no.'s when typeofdata is NN
For instance a field with a typeofdata of `N~O` will be formatted to users specs, but `NN~O` will not.

Luke1982:


joebordes:
thanks

PR #716, opened 2020-02-01
Fix regular main search

### LINKED ISSUE #702: Main search doesn't work when other autcompletes are active
This seems to be the case when you search something that the main search could not process into an autocomplete suggestion

Luke1982:
Fixes #702

### DIFF
--- a/include/js/general.js
+++ b/include/js/general.js
@@ -5314,6 +5314,9 @@ AutocompleteRelation.prototype.get = function (e) {
 				var json_data = JSON.parse(r.response);
 				if (json_data.length == 0) {
 					acInstance.clearTargetUL();
+					if (!!window.currentAc) {
+						window.currentAc.deactivate();
+					}
 				} else {
 					acInstance.set(json_data);
 				}
--- a/service-worker.js
+++ b/service-worker.js
@@ -9200,7 +9200,7 @@ if (workbox) {
   },
   {
     "url": "include/js/general.js",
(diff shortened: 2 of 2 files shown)


Luke1982:
Travis failed on the double boolean casting, don't know if that's really a problem?

joebordes:
I would prefer to eliminate that double boolean. I really don't understand what it is doing but I can imagine that it is some edge functionality of javascript that converts the value to boolean in case it isn't and that may break two years from now.
It makes it harder to read and maintain the code as you are presuming some advanced knowledge of JS that many may not have.
Can you change the condition to something more understandable?

joebordes:
manually merged this one since I had to update the service worker also and I tried to convert the double boolean to an equivalent conditional with no sucess: JS is a mess

Thanks!

Luke1982:
I love the double boolean, it's simply a way of saying 'is anything but false'. I don't know the implications of the perils that could happen, but maybe there's something I don't know yet. Anyway, you could just replace it with
```javascript
if (variable !== false)
```

Luke1982:
> manually merged this one since I had to update the service worker also and I tried to convert the double boolean to an equivalent conditional with no sucess: JS is a mess
> 
> Thanks!

About the serviceworker, I updated it but you'd like it in a different maner?

joebordes:
you updated the service worker correctly, it was just that I had to update it also due to the fix of moving the inventory lines and since I was already testing your change I decided to update your change and mine in one commit

this PR was totally correct: no worries

PR #720, opened 2020-02-18
feat (GlobalVariables): Define a Global Variable for Inventory Quanti…

emukaevo:
…ty Precision


joebordes:
manually merged.
thanks!

PR #725, opened 2020-03-04
feat(ModTracker) Start implementing LDS and Toast UI

Luke1982:
OK nowhere near ready but this works. **Don't forget** to add Business Actions for the Toast CSS and JS file. What we need to do at least still:

- Create an update that'll create the BA records for the Toast CSS and JS
- Perform some action when a user is not permitted to see a certain 'view'
- Perform some action when a user is not permitted to see changes for a certain field
- Show that the record was deleted
- Show that the record was restored
- Translate all the hardcoded strings
- The highlighting. Not sure why that's there, since you clearly see the old and the new value so why highlight it?

When we're all done, we can delete the Smarty templates associated with the ModTracker


joebordes:
I'll give it a try as soon as I can but a few things already:

- shouldn't we extract those "javascript components" into their own files?
- the highlight functionality works on text fields, it shows you the text that has changed. In most others (numeric, picklist) it doesn't give you any more information.

joebordes:
This looks great.

> Create an update that'll create the BA records for the Toast CSS and JS
I added a mechanism to load components and loaded toast directly in the header. Although it won't be needed by all pages it will be needed by many, so, for the moment everywhere. I added it to the service worker

> Perform some action when a user is not permitted to see a certain 'view'
I added the check for permission

> Perform some action when a user is not permitted to see changes for a certain field
I did not look into this case. Wasn't it directly supported before the graphical change?

> Show that the record was deleted
> Show that the record was restored
I vaguely remember having tried this and implemented something for it. Have you tried it?

> Translate all the hardcoded strings
Done. I also added a mechanism to load translation strings

> The highlighting. Not sure why that's there since you clearly see the old and the new value so why highlight it?
I made some changes here. Activated it always but only for some fields and I fixed some other field types.

Open new tickets for the pending issues if you can

Thanks!!

Luke1982:
Hey that's cool!

About the 'moving to own files', that's a tricky one, since I think I read somewhere that you should do that for webcomponents. This isn't a true WC since it doesn't extend HTMLElement and doesn't implement the other WC rules.

So yeah, you *could* move the prompt and modal to separate files, but I don't think the 'modern way of doing it' compells you to. But, if the code will grow (which it surely shall as we add more functionality) it may still prove to be a good idea.

joebordes:
I did that. I separated the code into files. No modules nor class, just "as is" but each "component" in its own file

Luke1982:
Yeah I saw, better choice with our future selves in mind. Where do the translations live? So I can do a PR for the Dutch ones.

joebordes:
they live in the modules' language directory nl_nl.js

Issue #727, opened 2020-03-08
Tracker needs two clicks at endpoint, should react right away


Luke1982:
https://github.com/tsolucio/corebos/blob/59c0cd002f30a40bb9557ad24db41af14d50a71d/modules/ModTracker/ModTrackerCommon.js#L111-L117

When the tracker has no more history to show in any direction, it shows a message. But you need to click twice before you get the message in stead of once. Should improve that for better UX.

Luke1982:
Hm, this is not a problem with the new tracker implementation. It seems that multiple 'atpoints' produce the same result sometimes. I never got into how the tracker works but it seems it can store duplicate entries sometimes. Not a frontend issue but I think very hard to find...

joebordes:
maybe just a test-data issue. When I created the coreBOSTest database I did not always follow the rules. I did some direct inserts and duplicated information simply because it was easier.

Let's close this and open an issue if we do find some duplicates somewhere

Luke1982:
Nah it does that in a regular installation as well, but again, not a big deal. Not worth the time right now

PR #731, opened 2020-03-11
feat(ldsPrompt) Apply shadow to entire prompt as opposed to only footer

Luke1982:
Originally I wanted to tackle https://github.com/tsolucio/corebos/issues/727, but it seems that will prove more difficult than I anticipated.


joebordes:
thanks!

PR #734, opened 2020-03-16
fix(Settings) Setting and Unsetting Related List Module for UIType 10  field in Module Layout Editor

mocodertz:


modules/Settings/LayoutBlockList.php
```diff
@@ -711,6 +711,9 @@ function updateFieldProperties() {
 				$parentmodule = Vtiger_Module::getInstance($fld_module);
 				$relationfield = Vtiger_Field::getInstance($fieldname, $parentmodule);
 				$relationfield->setRelatedModules($module);
+				$relatedModule = Vtiger_Module::getInstance($module);
+				$parentmodule->setRelatedList($relatedModule, '', array('ADD','SELECT'), 'get_related_list');
```
joebordes: The label has to be the related module name
Here we are adding uitype 10 capture fields. That means that we are setting 1:M relations, so you have to use get_dependents_list not get_related_list.

joebordes:
thanks!

PR #751, opened 2020-04-02
Improve View of ModTracker Entries

denaldd:


denaldd:
We need to make this change in BusinessActions if we need to add the complete=1:
**javascript:ModTrackerCommon.showhistory('$RECORD$', 0,false, true)**

joebordes:
thanks!

PR #754, opened 2020-04-07
Main Table Alias and CRM Entity Alias

george1806:


modules/cbQuestion/workflow/mview.php
```diff
@@ -32,7 +32,11 @@ function CBQuestionMViewFunction($entityData) {
 		list($void, $eid) = explode('x', $entityData->getId());
 		while ($cbq = $adb->fetch_array($qs)) {
 			$vname = str_replace(' ', '_', $cbq['qname']);
-			$adb->query('delete from '.$vname.' where '.$cbq['uniqueid'].'='.$eid);
+			if (!empty($cbq['maintablealias'])) {
+				$adb->query('delete from '.$vname.' where '.$cbq['maintablealias'].'.'.$cbq['uniqueid'].'='.$eid);
+			} else {
+				$adb->query('delete from '.$vname.' where '.$cbq['uniqueid'].'='.$eid);
+			}
```
joebordes, line 39: This is not needed as we are creating the whole SQL command and we know the name of the table we are deleting on.
undo this change, the maintablealias is only for the update SQL command we manipulate below.

joebordes:
Please move these fields also to the new advanced block:

- Unique ID Field
- Map Id
- Conditions in Filter Format
- Materialized View Cron
- Materialized View Workflow

You can use [massMoveFieldsToBlock](https://github.com/tsolucio/corebos/blob/master/modules/cbupdater/cbupdaterWorker.php#L470)

Add these translations to the es_es and es_mx files, please.

``` php
'LBL_cbQuestion_Advanced_Usage' => 'Opciones Avanzadas',
'Main Table Alias' => 'Alias de la Tabla Principal',
```

thanks

joebordes:
Since you are creating changeset to change the question fields, please add an SQL update to set the vtiger_field.helpinfo='SQLDELETE' for the sqlquery field 

thanks

modules/cbQuestion/workflow/mview.php
george1806: Okay

joebordes:
looks good!  Let's see if we got it right this time :-)

Thanks!

george1806:
Okay! lets see.. thanks

PR #758, opened 2020-04-16
fix(FieldBasic) Adding fieldname and check for both name and label

tebajanga:


joebordes:
looks great!!
thanks!

Issue #774, opened 2020-05-14
Pageheader needs finetuning


Luke1982:
Can be jittery sometimes

Luke1982:
This has been taken care of for now, let's open a new issue if we find something else.

PR #775, opened 2020-05-14
feat(PageHeader) Tweak movement, related to #774

Luke1982:
Wrote a new function that throttles scroll specifically, with the main addition that it does not only throttle time, but also scrolled distance. Did some further tweaking, set some timeouts. Should be a lot less jerky now.

### DIFF
--- a/include/js/general.js
+++ b/include/js/general.js
@@ -5166,6 +5166,29 @@ var throttle = function (func, limit) {
 	};
 };
 
+/*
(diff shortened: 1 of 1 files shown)


joebordes:
Did some very basic testing, it does seem better.
I'm going to merge it in and get back to you with comments when I can

Thanks!

Luke1982:
Cool, let me know what feedback you receive

PR #779, opened 2020-05-20
feat(cbMap) add xsd file for Information Map

glorymoshi:


modules/cbMap/XSD_schemas/InformationMap.xsd
```diff
@@ -0,0 +1,16 @@
+<xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="map">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="information">
+          <xs:complexType>
+            <xs:sequence>
+              <xs:element type="xs:string" name="infotype"/>
+              <xs:element type="xs:date" name="value" maxOccurs="unbounded" minOccurs="0"/>
```
joebordes, line 9: this has to be of type "string" 
the example in the code is an array of dates, but other maps could contain any type of information

joebordes:
thanks!

PR #782, opened 2020-05-27
i18n(Module) translate module labels to italian

elisadeko:


joebordes:
I merged these translations. better to have them in the application than here waiting and you can continue as time permits by creating other PRs

Thanks!

PR #783, opened 2020-05-29
fix(Invoice) return a lost product line in potential module

glorymoshi:


denaldhushi:
@joebordes only to Potentials module was the problem.

joebordes:
great!  thanks!

PR #784, opened 2020-05-29
Converting  the form fields for VT Entity Method Task to LDS

rdsande:
Converting the fields, removing the old table tags and substituting them with the div and classes from the LDS system


Smarty/templates/com_vtiger_workflow/EditTask.tpl
```diff
@@ -28,84 +28,191 @@
...
-
-		<table class="tableHeading" border="0"  width="100%" cellspacing="0" cellpadding="5">
-			<tr>
-				<td class="big" nowrap="nowrap">
-					<strong>{$MOD.LBL_TASK_OPERATIONS}</strong>
-				</td>
-			</tr>
-		</table>
+		</div>
+
```
joebordes, line 214: please eliminate one of these two empty lines

Smarty/templates/com_vtiger_workflow/taskforms/VTEntityMethodTask.tpl
```diff
@@ -8,17 +8,30 @@
  * All Rights Reserved.
  ************************************************************************************/
 -->*}
+
```
joebordes, line 11: in the future do not add empty lines like this

Smarty/templates/com_vtiger_workflow/EditTask.tpl
rdsande: Ok

Smarty/templates/com_vtiger_workflow/taskforms/VTEntityMethodTask.tpl
rdsande: Alright

PR #785, opened 2020-06-02
Lds advanced filter

Luke1982:
This isn't completely done yet, haven't tested reporting. I've branched off the ListViewJSON branch that Denald created so that part is already integrated. We probably want to create a new PR in the end, this one is mainly good for keeping the commits and changes recorded.


Luke1982:
OK, this is a first working version now:
- Listview search works
- Customview works
- Reporting works both as a modal the full report

There are bound to be some quirks and bugs though. Please review and let me know so we can fix them and merge before the master gets too different. I've merged the latest commits from ListViewJSON so we are on track with that branch.

Luke1982:
Fixed a bug where I used translated values for the gluevalues instead of just 'and' and 'or'

joebordes:
This PR is installed on the denormlv branch where I have been testing it. Seems to be working correctly in general but I ran into an issue today in the Popup where it isn't working, I fixed a couple of warning by missing variables (which I may have done incorrectly)

https://github.com/tsolucio/corebos/commit/217d15c68893d51a3abb7304ea785d830da22b4e

LDSAdvancedFilterPopupError

@Luke1982 can you fix that?

Luke1982:
@joebordes Your changes seem correct and match the ones I made. But it is not the full solution. Also, as mentioned, we need to asses if we need the original options retrieval function in the listview controller

joebordes:
fixed with PR #984 
Let's keep testing...

Luke1982:
Keep me posted if you run into problems. Would be very nice if we could almost get the entire listview adapted to a modern UI

joebordes:
I am pretty confident of releasing the denormlv branch as an official release for the new year

Would like some more testing but I think that is going to happen very soon

Luke1982:
I'm not sure how I can test. It would help to know what you've tested so far so I can asses if there are usecases I can think of that you haven't tested yet

joebordes:
manually merged into master
thanks!

PR #787, opened 2020-06-04
Fix/rds vt entity method edit task

rdsande:
Converting VTEntityMethodTask and EditTask to LDS also applying some width and z-index to the number box in the style file.


joebordes:
Thanks!

PR #788, opened 2020-06-08
Converting the VTEmailTask taskform to LDS

rdsande:


joebordes:
looks great. good work
thanks!

PR #791, opened 2020-06-23
feat(ListView) add ListView Component

denaldhushi:


Luke1982:
I'm following this PR since I have another one open that updates the advanced filtering and I want to merge those changes in right away. I've noticed that the inline edit doesn't work well yet when it comes to dropdowns. Check the console, it'll give you an error. Apart from that, great work!

denaldhushi:
hi @Luke1982. Thank you!
I looked into the error, and looks like is the problem with `renderer` and `editor` at column header. I opened an issue about this [here](https://github.com/nhn/tui.grid/issues/1104)

Luke1982:
Hi @denaldhushi OK I had the same thing when I tested TUI grid, immediately ran into some bugs. The team does fix them when you submit reports, may take a while though.

denaldhushi:
Issue with dropdown is fixed right now and grid is updated to the latest release **v4.14.0**.

joebordes:
Hi,

quick question. I just accepted [this PR](https://github.com/tsolucio/corebos/pull/793) and thought: have you tried this functionality on the new data table? is it still working?

Luke1982:
> Issue with dropdown is fixed right now and grid is updated to the latest release **v4.14.0**.

Nice!

@joebordes How are we doing on the acceptance roadmap here? I'd like to get on with advanced filtering and then move to detailview LDS, but don't think it would be wise to work on detailview while this PR and the advancedfilter one are still open.

Luke1982:
> Hi @joebordes
> Tooltip functionality is not working with new data table. We have to add this functionality also.

Would this be *a lot* of work? Implementing the old-style tooltip? Because maybe we're better off renewing that one as well, using LDS while we're at it.

joebordes:
Incredible work. really!!

I will give this another review and see if we can get it into coreBOS ASP.

thanks!!

Luke1982:
I would like to offer [this](https://github.com/Luke1982/corebos/commit/ff430b24aeaf70213913fa55dfb72a6292d14d2f) commit, that moves the tooltip functionality to CSS. One thing I would like to discuss is [this](https://github.com/Luke1982/corebos/commit/ff430b24aeaf70213913fa55dfb72a6292d14d2f#diff-76b5fa3f673f0ba775240ef8f40cfe2dR113) of which I'm not sure it won't break anything.

joebordes:
that looks very nice @Luke1982 
I think I would activate it only on the information icon. The whole-cell is how we do it now and it seems rather intrusive. Adding the icon is the perfect excuse to show it only when you ask for it.
As to the change you point to, I don't know what the potential danger could be.

Luke1982:
@joebordes I agree, the 'i' is commonly used and much less invasive. Let me whip up some code

Luke1982:
OK, check [this](https://github.com/Luke1982/corebos/commit/0c5feef4404867f20f07e20bff320b0d135f01c9) commit that improves the tooltip triggering.
Peek 2020-07-21 10-23

Luke1982:
So @denaldhushi I think [this](https://github.com/Luke1982/corebos/commit/ff430b24aeaf70213913fa55dfb72a6292d14d2f#diff-382791634de0855bc76ca9a85bae4ee8R186-R189) CSS I wrote should be moved to a better location and you should review [this](https://github.com/Luke1982/corebos/commit/ff430b24aeaf70213913fa55dfb72a6292d14d2f#diff-76b5fa3f673f0ba775240ef8f40cfe2dR113) change to see if it breaks something you designed.

denaldhushi:
@Luke1982 maybe we can move the CSS [here](https://github.com/tsolucio/corebos/pull/791/files#diff-6a2a49fc3237e6f118a01f5042bf20d2R44)
I have reviewed the change about `innerHTML` and looks like is correct.

Luke1982:
@denaldhushi Hmm yes then best place it in 'include/components/listview/listview.css' yes. Great, will you manage to copy the two commits I made into your branch, to improve the tooltip on the PR?

denaldhushi:
okay @Luke1982 I will update this branch with your changes about the tooltip

Luke1982:
@denaldhushi Cool, let me know what you think and if you see room for improvement. The lineheight should be improved, but I think that is better suited in an overall LDS PR where we use the correct font and font-size.

Luke1982:
Guys, how are we on this? I feel like there's not much movement, is there some hurdle we still need to take?

joebordes:
we are working on #214 which goes first. Almost there. After that one we will do this one

PR #808, opened 2020-07-17
fix() Enhance AND and OR to accept more than two parameters

george1806:


Luke1982:
Now this is a nice feature, one that I miss. Would like to start documenting vtlib expression syntax. @joebordes could you help me on my way?

joebordes:
thanks!

joebordes:
@Luke1982 absolutely

this has been on my endless to-do list for ages.

My initial idea was to create a blog post series explaining the workflow system, that would take me to the expression language and there I wanted to add documentation to the wiki and keep blogging

But that doesn't seem to be happening. What is happening is that we keep getting new functions and more functionality. I have even discussed with Denald to review [the list of spreadsheet functions](https://support.microsoft.com/en-us/office/excel-functions-alphabetical-b3944572-255d-4efb-bb96-c6d90033e188?ui=en-us&rs=en-us&ad=us) and add more.

So now, what I have in mind is to create a "help" section inside the expression editor. Enhance the function selection. Change it from a "now hard to read select box" into something else. Some screen where you can see the full list of functions, filter per category and get a small help/information/explanation of what and how each function works.

With that approach, I expect to "obligate" us to create documentation as we add the functions, make it multilingual and keep it up to date easier. Besides the obvious benefit of making the expression editor easier to use, more user-friendly.

Obviously, the blog and wiki documentation are still something that has to be done too.

How does that sound?

Luke1982:
Sounds like a good plan because each commit that adds functionality requires you to document it right away. We can use the LDSmodal component to show that in, some datatable to clearly format function and arguments, and so on.

joebordes:
I have finished the first version of this functionality and just released it

Luke1982:
Nice!

Issue #811, opened 2020-07-25
Adding a related list for users produces a broken query


Luke1982:
If you add a related list that uses `get_related_list` and relate your module to the 'Users' module, it won't work since the query produced will look something like:
```sql
SELECT count(*) AS count 
	FROM vtiger_users 
	INNER JOIN vtiger_crmentity ON vtiger_crmentity.crmid = vtiger_users.id 
	INNER JOIN vtiger_crmentityrel ON (vtiger_crmentityrel.relcrmid = vtiger_crmentity.crmid OR vtiger_crmentityrel.crmid = vtiger_crmentity.crmid) 
	LEFT JOIN vtiger_users ON vtiger_users.id = vtiger_crmentity.smownerid 
	LEFT JOIN vtiger_groups ON vtiger_groups.groupid = vtiger_crmentity.smownerid 
	WHERE vtiger_crmentity.deleted = 0 
		AND (
			vtiger_crmentityrel.crmid = 74 
			OR vtiger_crmentityrel.relcrmid = 74
		)
```
That fails since the `vtiger_users` table is joined on itself without aliasing it (see the first LEFT JOIN). It should be:
```sql
SELECT count(*) AS count 
	FROM vtiger_users 
	INNER JOIN vtiger_crmentity ON vtiger_crmentity.crmid = vtiger_users.id 
	INNER JOIN vtiger_crmentityrel ON (vtiger_crmentityrel.relcrmid = vtiger_crmentity.crmid OR vtiger_crmentityrel.crmid = vtiger_crmentity.crmid) 
	LEFT JOIN vtiger_users AS users2 ON users2.id = vtiger_crmentity.smownerid 
	LEFT JOIN vtiger_groups ON vtiger_groups.groupid = vtiger_crmentity.smownerid 
	WHERE vtiger_crmentity.deleted = 0 
		AND (
			vtiger_crmentityrel.crmid = 74 
			OR vtiger_crmentityrel.relcrmid = 74
		)
```

joebordes:
added with 70a5d68afc90da4b802dd7d908dc74b39ab10ce2

PR #812, opened 2020-07-25
fix(Users) Enable 'get_related_list' for Users by aliasing hardcoded …

Luke1982:
…join, related to #811

### DIFF
--- a/data/CRMEntity.php
+++ b/data/CRMEntity.php
@@ -2453,7 +2453,7 @@ class CRMEntity {
 		$query .= " INNER JOIN vtiger_crmentity ON vtiger_crmentity.crmid = $other->table_name.$other->table_index";
 		$query .= ' INNER JOIN vtiger_crmentityrel ON (vtiger_crmentityrel.relcrmid = vtiger_crmentity.crmid OR vtiger_crmentityrel.crmid = vtiger_crmentity.crmid)';
 		$query .= $more_relation;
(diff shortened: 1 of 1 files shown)


joebordes:
In the past, when I have had to implement this feature (twice if I remember correctly) I ended up implementing a custom related list function or a new module to hold the relation (project stakeholders for example).

That said, I see what you are trying to do here by reusing the vtiger_crmentityrel table and existing functionality and I think it is a good idea, but your solution is insufficient/incorrect.

The "left join" that you applied the alias to has no meaning at all in the case of the user module. There is no smownerid field in the vtiger_crmentity table to join on, so that "left join" and the next one for the group will always be empty, useless. I would prefer that we create a more optimized SQL, adding an "if" to see if we are joining on Users and, in that case, leave out the two left joins completely and also change the "vtiger_crmentity.deleted" to "vtiger_users.deleted" which is the correct column to check that state on.

Finally, to really implement this functionality completely, we should add the Users module to the "Add Related List" dropdown in the layout editor.

Let me know what you think

joebordes:
another idea would be to add a new method: get_related_user_list() specifically for Users, similar to get_emails and get_attachments....

Luke1982:
I would opt for doing it the first way you mentioned. Although it's more work, I am a proponent of standardizing this sort of stuff into functionality we already have in stead of adding functions and code for very specific use cases. Let me take your comment and see what I can do.

PR #816, opened 2020-07-28
fix(LDS) Remove border on firefox search icon, fixes #749

### LINKED ISSUE #749: Search icon in firefox looks chopped
At first I though this was an LDS issue, but upon using LDS in a non-coreBOS project I noticed that it was behaving correctly there. There must be something somewhere in our stylesheets that is throwing it off.

Luke1982:


joebordes:
thanks!

PR #824, opened 2020-07-31
Cleanup

jsanga1j:


Smarty/templates/ActivityReminderCallback.tpl
```diff
@@ -15,7 +15,7 @@
 	<tr>
 		<td align='left'><b>
 		<img align="top" src="themes/images/{$activityimage}s.gif"/> {$activitytype} - {$cbstatus}</b> </td>
-		<td align='right'><b><font color={$cbcolor}>{$cbdate} {$cbtime}</font></b></td>
+		<td align='right'><b><div color={$cbcolor}>{$cbdate} {$cbtime}</div></b></td>
```
joebordes, line 18: this doesn't look correct, shouldn't it be

`<span style="color:{$cbcolor}">`

?

Smarty/templates/ComposeEmail.tpl
```diff
@@ -56,7 +56,7 @@
 	</td>
 </tr>
 	<tr>
-	<td class="mailSubHeader" align="right"><font color="red">*</font><b>{$MOD.LBL_FROM}</b></td>
+	<td class="mailSubHeader" align="right"><div style="color: red;">*</div><b>{$MOD.LBL_FROM}</b></td>
```
joebordes, line 59: can we follow LDS recommendations for this. Instead of just changing it, why don't we apply LDS now and avoid having to change it all again later?

joebordes:
there are too many files changed in this PR. I am going to give you a list of files you need to change.

joebordes:
closing

PR #830, opened 2020-08-05
fix(PurchaseOrders) Don't hardcode vendor address when vendor is pres…

Luke1982:
…ent in request


joebordes:
How can we do this without losing the current functionality?

Luke1982:
You can't, except when you install a businessmap that performs the old behavior in the same update. But that can't be a reason to keep something that is so fundamentally wrong. Ordering a product at a vendor and telling them to ship and bill it to themselves is something that I don't believe any company will do.

joebordes:
fixed with d1466712398122888622f6709d66c6d8247a75b5

thanks!

PR #833, opened 2020-08-06
feat(ProductAutocomplete) Prepare autocomplete to work with modern in…

Luke1982:
…ventoryblock


joebordes:
I can't accept this PR
Mainly because it is based on an outdated master branch, so it took me WAY to long to get to a point where I actually saw some code change. I would ask you to

- always start your branches off the latest master branch: pull before creating the branch
- if you have to do a massive indentation, do two commits, one that has the indentation and that i can basically ignore and another with the significant change you are trying to accomplish

When I finally arrived to see a significant change I see that you eliminate the multicurrency code, why would you do that?

Please create a new PR based on the latest branch and explain why we eliminate the multicurrency code.

PR #834, opened 2020-08-10
feat(Smarty) implement LDS combobox support on uitype 1

shibe2018:


Luke1982:
Hi @shibe2018. I don't think this is the best way to implement the ldsCombobox. I wrote an extensive [JS component](https://github.com/tsolucio/corebos/blob/master/include/js/general.js#L5701) that does exactly this and your code seems to duplicate a lot of that functionality (like in the [global search](https://github.com/tsolucio/corebos/blob/master/include/js/general.js#L5077). What is it exactly that you're trying to accomplish? I think it would be better if we try to implement that functionality by using the existing code, which I intentionally wrote to be futureproof and multifunctional.

Smarty/templates/EditViewUI.tpl
```diff
@@ -240,11 +240,112 @@ alt="{'LBL_CLEAR'|@getTranslatedString}" title="{'LBL_CLEAR'|@getTranslatedStrin
...
-			<td id="td_{$fldname}" width=20% class="dvtCellLabel{if $mandatory_field == '*'} mandatory_field_label{/if}" align=right><font color="red">{$mandatory_field}</font>{$usefldlabel} {if $MASS_EDIT eq '1'}<input type="checkbox" name="{$fldname}_mass_edit_check" id="{$fldname}_mass_edit_check" class="small">{/if}</td>
-			<td id="td_val_{$fldname}" width=30% align=left class="dvtCellInfo"><span style='display:none;' id='{$fldname}_hidden'></span><input readonly type="text" tabindex="{$vt_tab}" name="{$fldname}" id ="{$fldname}" {if $MODE eq 'edit'} value="{$fldvalue}" {else} value="{$MOD_SEQ_ID}" {/if} class=detailedViewTextBox onFocus="this.className='detailedViewTextBoxOn'" onBlur="this.className='detailedViewTextBox'"></td>
-		{elseif $uitype eq 11 || $uitype eq 1 || $uitype eq 13 || $uitype eq 7}
+				<td id="td_{$fldname}" width=20% class="dvtCellLabel{if $mandatory_field == '*'} mandatory_field_label{/if}" align=right><font color="red">{$mandatory_field}</font>{$usefldlabel} {if $MASS_EDIT eq '1'}<input type="checkbox" name="{$fldname}_mass_edit_check" id="{$fldname}_mass_edit_check" class="small">{/if}</td>
```
joebordes, line 243: please undo this indentation. you should do a `git diff` BEFORE committing and review each line, you should commit only relevant lines of changes

joebordes:
@shibe2018 

- you should put javascript code in a javascript file, not in the template. in a file, it will get included in cache instead of generated and sent each time the template is processed (at least as much as possible)
- I agree with @Luke1982, we should use existing code as much as possible. when you see that you have to implement something that is already implemented somewhere else you have to read that code and see how it works, in order to use it or enhance it to your needs. His code is already in some cached JS file so you get that directly
- I see that you add the functions you need directly in the template: what happens if there is more than one field with this functionality. The function will be duplicated on the end page. Will they work? Please try with more than one field configured

@Luke1982 he is trying to implement [this functionality](https://corebos.com/documentation/doku.php?id=en:adminmanual:businessmappings:fieldinfo:combobox). Basically a normal uitype 1 text field where you can put any value you want but you can also choose from a pre-defined set of "typical" values. So it is like picklist but with the option to put any other value if you need to

Can you please help him? Point him to your code and how it works. See what would be needed to implement the desired functionality

Luke1982:
@joebordes Sure. First of all you need to create a markup that resembles [this](https://github.com/tsolucio/corebos/blob/master/Smarty/templates/Header.tpl#L132-L145) markup, which is of course following LDS. Then you can do something like [here](https://github.com/tsolucio/corebos/blob/master/include/js/general.js#L5077) where you set some params (most notably, you can pass a custom callback, but in this case I wouldn't) but I don't think you need any, other when maybe explicitly setting 'isMulti' to false. That parameter was created to allow comboboxes to select multiple value like the one in the header next to the global search.

Basically The combobox already does a lot of what you want. It just doesn't save the values yet. To mimic the behavior of an HTML `<select>` I introduced the concept of a **valueholder**, which is by default a hidden input that holds the actual value to be sent to the backend on saving. Typically, a select's `<option>` has a label it shows the user and a value it actually saves.

You specify the valueholder by a data attribute on the main [input](https://github.com/tsolucio/corebos/blob/master/include/js/general.js#L5811-L5836). I would suggest adding a new parameter to the combobox param object, something like 'updateOnType' that allows the 'keyup' event to also update the valueholder (by default that only happens when selecting one of the options). That would get us a long way in the right direction I believe.

shibe2018:
Hello
Thanks for support am on checking luke script and try to adapt the functionality

Smarty/templates/EditViewUI.tpl
shibe2018: yes sure i see let me change this

Luke1982:
@shibe2018 No problem, keep me informed.

Also, please try and keep the function bodies small. It is likely that we have to extend the combobox in the future and the smaller the function bodies are (try to split functionality into small conceptual units) the easier and more efficient the component will be.

shibe2018:
I understand Joe I will do that and I will do my best to finish this on time

PR #836, opened 2020-08-13
feat(ProductAutocomplete) Prepare for modern inventory block

Luke1982:
OK, so here's a new one off an up-to-date master and without indentation changes. As I explained in a PR that I closed before it had a change to be reviewed, the multicurrency is not gone, you just placed it in the wrong place. The `fillLine` method of the module uses the '[else](https://github.com/tsolucio/corebos/pull/836/files#diff-dc52855ef3d2e5662b79a6512e087a12R1804)' to do a custom callback in the current inventorymodules. In this case hand the information over to [`handleProductAutocompleteSelect`](https://github.com/tsolucio/corebos/blob/master/include/js/Inventory.js#L1938), which is where the multicurrency happens.

The main block of `fillLine` uses the [modules](https://github.com/Luke1982/WorkAssignment/blob/master/modules/WorkAssignment/WorkAssignment.js) I wrote to handle modern inventorylines, which were originally written to work together. I only took out the autocomplete and suited it for the current way of doing inventory so we wouldn't have to wait so long on autocomplete in the product search fields.

### DIFF
--- a/include/js/Inventory.js
+++ b/include/js/Inventory.js
@@ -1420,10 +1420,10 @@ function InventorySelectAll(mod, image_pth) {
 }
 
 /****
-* ProductAutocomplete
-* @author: MajorLabel <info@majorlabel.nl>
-* @license VPL
-*/
+	* ProductAutocomplete
+	* @author: MajorLabel <info@majorlabel.nl>
+	* @license VPL
+	*/
 (function productautocompleteModule(factory) {
 
 	if (typeof define === 'function' && define.amd) {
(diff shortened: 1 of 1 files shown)


Luke1982:
Hmm crap I see there still are indentation differences, let me fix that

joebordes:
thanks!

Luke1982:
OK, so now your installation should be ready (could be missing some CSS) to use the [WorkAssignment](https://github.com/Luke1982/WorkAssignment) module. That's far from ready but it does do the basics.

PR #840, opened 2020-08-14
fix(ProductAutocomplete) Prevent submitting on enter when the autocom…

Luke1982:
…plete is open and enter is used to select an option


joebordes:
thanks!

PR #841, opened 2020-08-14
feat(CRMEntity) Allow for users to be related through regular related…

Luke1982:
… lists


Luke1982:
As requested a better way to optimize the related query for users. What has not yet been implemented is the extra option for users in the layout editor

joebordes:
looks, ok. I'll give it a try. Thanks!

PR #858, opened 2020-09-03
Implement combobox feature at uitype1 fields

shibe2018:


modules/Vtiger/ExecuteFunctions.php
```diff
@@ -95,6 +95,29 @@
 			}
 		}
 		break;
+	case 'getFieldValuesFromMap':
+		$sourceModule = $adb->sql_escape_string(vtlib_purify($_REQUEST['sourceModule']));
+		$inputVal = $adb->sql_escape_string(vtlib_purify($_REQUEST['inputVal']));
+		$fieldname = $adb->sql_escape_string(vtlib_purify($_REQUEST['fieldname']));
+		if (!empty($inputVal)) {
+			$bmapname = $sourceModule . '_FieldInfo';
+			$cbMapid = GlobalVariable::getVariable('BusinessMapping_FieldInfo', cbMap::getMapIdByName($bmapname), $sourceModule, $current_user->id);
```
joebordes, line 104: the GV name here must be 'BusinessMapping_'.$bmapname

`GlobalVariable::getVariable('BusinessMapping_'.$bmapname, cbMap::getMapIdByName($bmapname),`

shibe2018: sure

modules/Vtiger/ExecuteFunctions.php
```diff
@@ -95,6 +95,29 @@
...
+			if (strlen($inputVal) > 0) {
```
joebordes, line 105: this condition is unnecessary as you have an `if (!empty($inputVal))` on line 102. If the execution gets here, the length will be bigger than 0 always

PR #859, opened 2020-09-03
Add soundex condition to querygenerator

shibe2018:


include/QueryGenerator/QueryGenerator.php
```diff
@@ -1535,6 +1547,11 @@ private function isDateType($type) {
 		return ($type == 'date' || $type == 'datetime');
 	}
 
+	private function requiresSoundex($operator) {
+		$requiresSX = array('sx');
+		return in_array($operator, $requiresSX);
```
joebordes, line 1552: just one line, directly return the result: `return in_array($operator, array('sx'));` there is no need to save the array in memory just to throw it away

joebordes:
this looks a lot easier than I thought it would be !!  Nice

Can you please add some unit tests on QueryGenerator with this operator and run the existing unit tests to make sure we don't break anything?

joebordes:
thanks!

PR #864, opened 2020-09-11
Implement regexp operation

shibe2018:


joebordes:
merged with soundex. thanks!!

PR #867, opened 2020-09-15
fix(Inventory) Avoid losing inventorylines when saving through save m…

Luke1982:
…ethod


joebordes:
perfect. tests failed before applying the change and passed after. I checked that totalProductCount is present when saving correctly from inside the application in full edit and from webservice while not present from inline edit 

thanks!!

Luke1982:
Cool! I think I personally have been bit by this a lot and I think we've really chopped this off at the root now.

PR #886, opened 2020-10-01
fix(Workflow) Checking referenceField for expression conditions

tebajanga:


joebordes:
works like magic.
I added two unit tests to control this case.
Thanks!

PR #897, opened 2020-10-07
feat(ldsCombobox) Allow starting the box disabled as readonly

Luke1982:
This doesn't allow changing the enabled state once the box is started but that could easily be added later, if we need it.


joebordes:
can you upload a screenshot of how it looks, please?

Thanks!!

Luke1982:
afbeelding
Looks just like it normally would, just doesn't respond to clicking or focussing. I wish LDS had a grayed out version but sadly it does not. We could remove the arrow icon on the right when the combo is not enabled

PR #910, opened 2020-10-12
fix(Inventory) Update price & taxes conditioned to dedicated function, fixes #909

### LINKED ISSUE #909: Product or service taxes are no longer saved
https://github.com/tsolucio/corebos/blob/fe15721f918d8a527a74e7c052ef3bf1e1c92802/modules/Products/Products.php#L121

When I fixed a bug [here](https://github.com/tsolucio/corebos/pull/867) I didn't realize the same function was used in Products and Services to determine if the taxes should be saved. That means that right now, no individual taxes can be saved at all. I believe that function should not be used for this purpose and that we should create a new function.

Luke1982:


Luke1982:
Can't believe something so simple happened. Let me try and create a 'good first issue' where I create a checklist.

PR #914, opened 2020-10-12
fix(InventoryDetails) Set Qty field as UI7, fixes #913

### LINKED ISSUE #913: InventoryDetails qty field is not set to UI7
When your code tries to format this field according to user preferences and your user is European, you'll end up with amounts multiplied by a 1000.

Luke1982:


Luke1982:
Could we take a look at this? It is causing problems for me when validating

joebordes:
manually merged 18cf59e775be84cbef56724d3071f61bd2ebadc5

thanks!

PR #917, opened 2020-10-13
feat(Utilities) Allow getting image information for a record through …

Luke1982:
…an AJAX call

### DIFF
--- a/modules/Vtiger/ExecuteFunctions.php
+++ b/modules/Vtiger/ExecuteFunctions.php
@@ -356,6 +356,17 @@ switch ($functiontocall) {
 	case 'getUserName':
 		$ret = getUserName(vtlib_purify($_REQUEST['userid']));
 		break;
+	case 'getImageInfoFor':
+		$id = vtlib_purify($_REQUEST['record']);
+		require_once 'include/Webservices/getRecordImages.php';
+		$imageinfo = cbws_getrecordimageinfo($id, $current_user);
+		header('Content-Type: application/json');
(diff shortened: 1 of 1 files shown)


Luke1982:
Example, the URL
`index.php?module=Utilities&action=UtilitiesAjax&file=ExecuteFunctions&functiontocall=getImageInfoFor&record=2617`
Will get you the image info on record 2617, if any

Luke1982:
I notice this only gets the latest image related to that record when you try and get images from products.

Luke1982:
@joebordes Could you take a look at this? I'd like it to return multiple images when multiple are available

joebordes:
well, the correct thing to do is add image fields on the module, not inside the special product multi-image field, but individual images with a name and a purpose. For example:

- home page image
- EAN image (which can be generated by a workflow)
- right side view
- left side view
...

not all bundled together inside one field

is that a solution for your use case?

Luke1982:
Well yes, that would be a much better way of doing things all together. Installations that do have multiple images on a product through the 'old' method would have to convert them to multiple fields, but that seems like a better way to do things so need break that egg to make the omelet I guess.

joebordes:
just to be clear: this works already, you don't have to do anything, getrecordimages returns all the image fields, the only field not supported is the special (non-standard) field in products which returns only one of the images in the field.

Luke1982:
Exactly. So if you want to get all the images for a product: they need to be on separate fields.

PR #937, opened 2020-10-28
Fix multi currency

Luke1982:
The collection of MC-prices from the DB is altered so that the actual price is used in stead of the unitprice, and the converted price is changed to the actual price when filling in the prices in the frontend. On top of that, during the filling of the prices there is a check that warns the user when:
- There is only one entry in the multicurrency object (indicating that basically there is no multicurrency)
- The actual price of that entry does not match the unit price (which should always be the same when there's only one currency active)

This warning is non-destructive: it'll just fill in the price from the MC object but will warn the user that something is up that needs checking. I left out the Spanish translation since you'd be better at that than Google Translate

### DIFF
--- a/include/js/Inventory.js
+++ b/include/js/Inventory.js
@@ -1976,7 +1976,10 @@ function handleProductAutocompleteSelect(obj) {
 	document.getElementById('comment'+no).innerHTML = obj.result.meta.comments;
 	var currency = document.getElementById('inventory_currency').value;
 	if (obj.result.pricing.multicurrency[currency] != undefined && gVTModule != 'PurchaseOrder' && gVTModule != 'Receiptcards') {
-		document.getElementById('listPrice'+no).value = obj.result.pricing.multicurrency[currency].converted_price;
+		if (Object.keys(obj.result.pricing.multicurrency).length == 1 && obj.result.pricing.multicurrency[currency].actual_price != obj.result.pricing.unit_price) {
+			ldsPrompt.show(alert_arr['Warning'], alert_arr.ACT_UNIT_PRICE_MISMATCH);
+		}
+		document.getElementById('listPrice'+no).value = obj.result.pricing.multicurrency[currency].actual_price;
 	} else {
 		var list_price = obj.result.pricing.unit_price;
 		if (gVTModule == 'PurchaseOrder' || gVTModule == 'Receiptcards' ) {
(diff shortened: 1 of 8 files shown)


include/js/Inventory.js
```diff
@@ -1976,7 +1976,10 @@ function handleProductAutocompleteSelect(obj) {
 	document.getElementById('comment'+no).innerHTML = obj.result.meta.comments;
 	var currency = document.getElementById('inventory_currency').value;
 	if (obj.result.pricing.multicurrency[currency] != undefined && gVTModule != 'PurchaseOrder' && gVTModule != 'Receiptcards') {
-		document.getElementById('listPrice'+no).value = obj.result.pricing.multicurrency[currency].converted_price;
+		if (Object.keys(obj.result.pricing.multicurrency).length == 1 && obj.result.pricing.multicurrency[currency].actual_price != obj.result.pricing.unit_price) {
```
joebordes, line 1979: is this check correct? shouldn't the actual price be different from the unit price for all currencies except the base currency?

If the base currency is EUR the unit price and the actual price in EUR should be the same, but the actual price in the other currencies may be different.  Am I wrong or is the "currency" variable always the base currency not the selected currency in the record?

Luke1982: Well I didn't write the MC code but yes, the `currency` variable is the currently selected currency, I just doublechecked that. 

The assumption I do is this: since we basically decided that **every** product has an entry in `productcrurrencyrel` for every currency it is active for and **every** product should at least have one currency (and that should be the base currency if it's only one), we can state that there will **always** be at least one entry in the `multicurrency` object that is returned from the database. I just checked that in the demo: you can't disable the base currency, it'll just re-enable that during the save.

That means that _when_ there is only one entry in that object (first part of the check), that should be equal to the `unit_price` (second part of the check).

Please verify and doublecheck this and let me know if you have questions

joebordes:
looks ok to me
thanks!!

Luke1982:
We've talked about how this approach will complicate saving the product unitprice through the code. How are we going to approach this? I mean: if someone creates a webservice where they update the product unitprice or use any other piece of code to update the unitprice, that will break the autocomplete in the app, since it won't reflect the latest price from that point on. I believe we have to do something about this, we can't risk users having incorrect data in their installations.

joebordes:
I learned yesterday that the retrieve method is incorrect also: in short neither web service nor the application code itself support multicurrency right now.

Luke1982:
This is not about not supporting multicurrency. That would be OK. But since we now treat every installation as multicurrency (since we use productcurrencyrel as leading), this is about not being able to update a product price, not sending that information to a user. You can't, as a developer, update the productprice through the code now, since the table's unitprice is not used and the multicurrency relation table is not updated when you update the unitprice. To do that, however, you have to get all existing currencies first and send them back unchanged since otherwise the application discards all of them. Again: in my opinion, this setup is a recipe for data corruption.

PR #947, opened 2020-11-03
Implement textarea customization option

shibe2018:


joebordes:
close in favor of #950

PR #951, opened 2020-11-06
feat(ProductServiceAC) Add compound searchfield support

Luke1982:
Also documented here https://gist.github.com/Luke1982/d886a67eb661db777d93e7e645076ecc#search-in-compounded-fields


joebordes:
Thanks!!

VERY good enhancement

PR #952, opened 2020-11-06
feat(ProductServiceAC) Add GV that allows tuning the searchlimit

Luke1982:


Luke1982:
Now that I'm thinking about it, this isn't the best way to do this. This does a GV lookup for every productline, while the result will always be the same. I think I'd rather move it outside of the ProductAC module and into the global scope.

joebordes:
the other GVs in javascript do it that way also, mostly because this is not immediate, there is a delay to get the value due to the ajax call. https://github.com/tsolucio/corebos/blob/master/include/js/Inventory.js#L14

joebordes:
and they set the default value in the "error"

Luke1982:
Yeah I should do a catch and set the global default var like that as well. Maybe I should leave it like it is, since I need to manipulate the query string on each fire (which is the keyup event) of the autocomplete, which may result in saving network calls on one side but trading them for higher memory use on the other. Need to let that sink in a little.

Luke1982:
@joebordes Please accept this as-is. If we move the globalvariable collection to the global scope, it won't have access to the autocomplete source strings, which it needs to alter the limit. The default limit is also already hardcoded in the AC source string, so no need to set a default in the error handler of the promise. If we were to move the GV collection to the global scope, we'd need to get that value and overwrite the source string of each AC on every keystroke.

joebordes:
Done.
Thanks!

PR #954, opened 2020-11-06
fix(ProductServiceAC) Don't show vendor and mfr part no when they're …

Luke1982:
…disabled

What the problem was: if you disabled either the vendor of manufacturer part no., you would still see them in the results.

### DIFF
--- a/include/Webservices/CustomerPortalWS.php
+++ b/include/Webservices/CustomerPortalWS.php
@@ -1069,9 +1069,15 @@ function getProductServiceAutocomplete($term, $returnfields = array(), $limit = 
 			vtiger_products.productname AS name,
 			vtiger_products.divisible AS divisible,
 			'Products' AS type,
-			vtiger_products.vendor_part_no AS ven_no,
+			CASE 
+				WHEN (SELECT presence FROM vtiger_field WHERE columnname = 'vendor_part_no' AND tablename = 'vtiger_products') = '1' THEN '##FIELDDISABLED##'
+				ELSE vtiger_products.vendor_part_no
+			END AS ven_no,
 			vtiger_products.cost_price AS cost_price,
-			vtiger_products.mfr_part_no AS mfr_no,
+			CASE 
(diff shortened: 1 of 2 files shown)


Luke1982:
What this does not solve is: when you don't setup a businessmap that controls the searchfields, the manufacturer and vendor part no.'s will still be searched in, even though they might not be active.

joebordes:
shouldn't we solve this by not adding the non-accessible fields to the $productsearchfields array?

``` PHP
	$productsearchfields = array('productname');
	if (getFieldVisibilityPermission('Products', $current_user->id, 'mfr_part_no') == '0') {
		$productsearchfields[] = 'mfr_part_no';
	}
	if (getFieldVisibilityPermission('Products', $current_user->id, 'vendor_part_no') == '0') {
		$productsearchfields[] = 'vendor_part_no';
	}
```

and then the implications that has in the rest of the function and code

Luke1982:
Well what I set out to do was try and prevent additional queries, or at least (since I do setup subqueries) prevent new DB connections. But yes I do think we should just prevent looking in the fields as well when they're deactivated since they are hardcoded fields when you don't override them with a business map. Let me try and rework the problem.

Luke1982:
@joebordes OK this is now preventing the fields from showing up under any circumstance, regardless of if the field in selected through a businessmap or not.

joebordes:
looks ok.
shouldn't we add `getFieldVisibilityPermission` on the fields in the Map also?

Luke1982:
I suppose so. We could walk through both products and services searchfields arrays (the default one is overwritten when you use a businessmap) and unset the keys based on the visibility/permission. For the part_no fields we should set the special `##FIELDDISABLED##` value so that the frontend knows to disable showing those fields.

In the future you'd want to customize the result box and do the `##FIELDDISABLED##` trick for all fields that are requested but not active or permitted.

joebordes:
thanks!

PR #958, opened 2020-11-10
feat(Smarty) Adding scroll bars to the notifications and last viewed …

shibe2018:
…panels


joebordes:
I made a change to hide the scrollbars because it wasn't looking good on smaller screens:

NPanelCSS

and moved the inline CSS to a class so it is easier to maintain and gets cached.

https://github.com/tsolucio/corebos/commit/887110e994c65d90595be6edf04b0aa6e6549f98

Thanks!!

PR #965, opened 2020-11-20
fix(cbQuestion) export results with Alias

denaldhushi:


modules/cbQuestion/actions/qactions.php
```diff
@@ -376,12 +376,16 @@ public function exportBuilderData() {
 		global $log;
 		$log->debug('> exportBuilderData');
 		$bqname = vtlib_purify($_REQUEST['bqname']);
+		$columns = json_decode(urldecode($_REQUEST['columns']), true)['headers'];
```
joebordes, line 379: please do this in two steps:

``` php
$columns = json_decode(urldecode($_REQUEST['columns']), true);
$columns = $columns['headers'];
```

I think that will have more support for older 7.x PHP versions and we still have a lot of those

joebordes:
thanks!

PR #974, opened 2020-12-02
fix(ProductAC) Send account/contact ID correctly when getting product…

### LINKED ISSUE #973: ProductAutocomplete doesn't send the account or contact ID
When searching in the Product autocomplete, the actual contact and account ID's should get sent in the request. In stead, strings with the values 'account_id' and 'contact_id' get sent.

Luke1982:
…/service, fixes #973


joebordes:
thanks!

PR #981, opened 2020-12-14
Implement delete user smtp settings

shibe2018:


include/integrations/smtp/settings.php
```diff
@@ -86,6 +86,23 @@
...
 	echo json_encode($response);
 } else {
+
+	$smtp_settings_mode = $_REQUEST['smtp_settings'];
+	/**
+	 * delete Incoming Mail Server Configuration
+	 */
+	if ($smtp_settings_mode == 'inc_set') {
+		$smtpconfig->clearInc_SMTPSettings();
+		header("Location: index.php?action=integration&module=Utilities&_op=getconfigsmtp&savemode=false");
```
joebordes, line 96: I think this is not needed. If you are in this "else" it is already getconfigsmtp, so you can just let it continue. Please delete it and try it to make sure it works the same with the redirect and without it. (in fact, I think it isn't working already, the menu HTML is breaking this header but you don't see the message because you don't have error reporting on, could be?)

include/integrations/smtp/settings.php
```diff
@@ -86,6 +86,23 @@
...
+	}
+	/**
+	 * delete Outgoing Mail Server Configuration
+	 */
+	if ($smtp_settings_mode == 'og_set') {
+		$smtpconfig->clearOg_SMTPSettings();
+		header("Location: index.php?action=integration&module=Utilities&_op=getconfigsmtp&savemode=false");
```
joebordes, line 103: same as the previous comment above

include/integrations/smtp/smtp.php
```diff
@@ -200,6 +200,29 @@ public function saveOutgoingMailServerConfiguration(
 		$adb->pquery($sql, $params);
 	}
 
+	public function clearInc_SMTPSettings() {
+		global $adb, $current_user;
+		$result = $adb->pquery('select 1 from vtiger_mail_accounts where user_id=?', array($current_user->id));
+		if ($adb->num_rows($result) > 0) {
```
joebordes, line 206: you don't need to make this check, if you launch an update and no record is found, MySQL will simply do nothing. So with one query you have enough. There is no need to change this, it isn't wrong to check nor important in this case as the performance impact is insignificant here. This comment is just for your knowledge.

include/integrations/smtp/smtp.php
```diff
@@ -200,6 +200,29 @@ public function saveOutgoingMailServerConfiguration(
...
+			$sql='update vtiger_mail_accounts set display_name=?, mail_id=?, mail_protocol=?, mail_username=?, mail_password=?, mail_servername=?,
+				box_refresh=?, mails_per_page=?, ssltype=? , sslmeth=?, status=? where og_server_status=0 or og_server_status=1 and user_id='.$current_user->id;
+			$params = array('
```
joebordes, line 209: eliminate this new line

include/integrations/smtp/settings.php
```diff
@@ -86,6 +86,23 @@
 	);
 	echo json_encode($response);
 } else {
+
```
joebordes, line 89: delete this empty line (coreBOS coding standard)

joebordes:
closing in favor of #983

PR #984, opened 2020-12-17
Repair new advanced filter

Luke1982:
Try these changes and see if they fix the problems you've been having


Luke1982:
Crap, you need to cherrypick them as I've merged all the development crap in as well

joebordes:
manually merged and seems to be working correctly. Thanks!

PR #985, opened 2020-12-18
feat(Product AC) Debounce and provide feedback on when search is perf…

Luke1982:
…ormed

I've tested this on a production server, works well. I chose a debounce in stead of a throttle since that would better mitigate the unwanted result of filling the resultbox with an outdated set of results


joebordes:
Thanks!
Looks good and a big plus for taking the time to format the HTML :-D

Luke1982:
Well I figured if we want to update somewhere soon this makes it much easier to do so. One thing I wonder about is the debounce delay. I used 600ms right now, would like to get some feedback about that, maybe it could be finetuned.

joebordes:
it does seem a bit "slow to react" on the validation I did. Maybe we could reduce it a little.

Luke1982:
https://github.com/tsolucio/corebos/pull/985/files#diff-c36fd6468bcfeb35da78043294b6e023822a5d1fb4790e8442e43dd554918fe7R1519

The second argument to the debounce here is the delay. Try moving that down while keeping the network tab open. My goal is to make it respond as quickly as possible while at the same time try to limit the number of requests to 1 per 'time you stop typing'

joebordes:
300 seems to be a good match on my computer, below that it calls too much. I think we could drop it to 400 safely.

Luke1982:
OK, I've set it to 400 and it seems to be a bit more responsive. It fires of the occasional second request per search but I think that is acceptable since it doesn't fill the resultlist with obsolete results anymore. If that does happen we need to make an extra check that prevents resultlists from rendering if a list with a newer timestamp is already rendered.

joebordes:
ok, I'll update it later today. thanks!

PR #986, opened 2020-12-28
add quite parameter to cron script

shibe2018:


vtigercron.php
```diff
@@ -22,12 +22,16 @@
 		// Run specific service
 		$srv = empty($argv[1]) ? $_REQUEST['service'] : $argv[1];
 		$srv = vtlib_purify($srv);
-		$srvcron = Vtiger_Cron::getInstance($srv);
-		if ($srvcron !== false) {
-			$cronTasks = array($srvcron);
+		if ($srv == '--quiet' || $srv == '-q') {
+			// do nothing
```
joebordes, line 26: I think you have to load all cron tasks here: `$cronTasks = Vtiger_Cron::listAllActiveInstances();` 
When we use the -q option we want to execute all the tasks without any messages, not ignore them all.

vtigercron.php
```diff
@@ -57,10 +61,12 @@
 					}
 				}
 				if (!$run_task) {
-					$msg = sprintf("[INFO]: %s - not ready to run as the time to run again is not completed\n", $cronTask->getName());
-					echo $msg;
-					$logbg->info($msg);
-					continue;
+					if (empty($argv[1]) || $argv[1] != '--quiet' || $argv[1] != '-q') {
```
joebordes, line 64: this condition is incorrect, I think you need parenthesis around the options, think about it and get back to me.

joebordes:
merged. give it a try when you can
thanks

PR #990, opened 2021-01-18
fix(Emails) check if a number is crmid

denaldhushi:


modules/Emails/mail.php
```diff
@@ -467,6 +467,23 @@ function addAttachment($mail, $filename, $record) {
...
+  * $id -- number or attachmentid
+  */
+function is_attachmentid($id) {
+	global $adb;
+	$adb->println('> is_attachment '.$id);
+	$result = $adb->pquery('SELECT attachmentsid FROM vtiger_attachments
+		INNER JOIN vtiger_crmentity ON attachmentsid=crmid WHERE deleted=0 AND attachmentsid=?', array(
+		$id
+	));
+	if ($adb->num_rows($result) > 0) {
```
joebordes, line 480: use a direct return here, it is a little faster:

`return ($adb->num_rows($result) > 0);`

you will have to move the "finished" debug message before the return

modules/Emails/mail.php
```diff
@@ -467,6 +467,23 @@ function addAttachment($mail, $filename, $record) {
 	}
 }
 
+/** Function to check if a number is a attachmentid
+  * $id -- number or attachmentid
+  */
+function is_attachmentid($id) {
```
joebordes, line 473: Should this function be in utils.php or CommonUtils.php in case we need to use it somewhere else? In fact, did you check if there is already a function that may do this?

joebordes: I had a look, couldn't find any function with this functionality. The closest one is [isRecordExists]8https://github.com/tsolucio/corebos/blob/master/include/utils/utils.php#L3312). I would suggest we move this function after (or before) that one.

joebordes:
Thanks!!

PR #999, opened 2021-02-25
Check for a dummy `health` key in the cache, in order to force throwi…

klivstudiosynthesis:
…ng exceptions for adapters that miss this behaviour on connection


joebordes:
thanks!

PR #1005, opened 2021-03-02
Implement add entityfield to module users

shibe2018:


modules/Users/Save.php
```diff
@@ -139,6 +139,9 @@
 	if (empty($focus->column_fields['roleid']) && !empty($_POST['user_role'])) {
 		$focus->column_fields['roleid'] = $_POST['user_role'];
 	}
+	$first_name = $focus->column_fields['first_name'];
```
joebordes, line 142: This is in the wrong place. If we create a user from a script or using the web service API, the field will not be filled in. Please move the code inside the save() function.

shibe2018: okay thanks

joebordes:
closing in favor of #1007

PR #1009, opened 2021-03-10
Implement monthday operator

shibe2018:


modules/com_vtiger_workflow/WorkFlowScheduler.php
```diff
@@ -422,6 +423,11 @@ private function _parseValueForDate($condition) {
 				$hours = $condition['value'];
 				$value = date('Y-m-d H:i:s', strtotime('-'.$hours.' hours'));
 				break;
+			case 'monthday':
+				$monthday = $condition['value'];
+				$value = date('m-d', strtotime($monthday));
+				var_dump($value);
```
joebordes, line 429: delete var_dump

joebordes:
merged. I had to do a lot of changes, most directly from the formatting and linting tools we have and while creating unit tests, but it was in the right direction.

thanks

PR #1012, opened 2021-03-11
Create a workflow Task that Send desktop Notifications

mocodertz:


omarllorens:
Hi.
I'll check the pull request and I've seen that you have an error on modules/cbupdater/cbupdates/2021.xml
When you add your changeset, you forgot to add an </changeset>

The rest of the code seems to be ok.

Can you fix it, please?

PR #1014, opened 2021-03-12
Optimize application sql changing user name concat for ename field

shibe2018:


modules/Products/Products.php
```diff
@@ -680,7 +680,7 @@ public function get_quotes($id, $cur_tab_id, $rel_tab_id, $actions = false) {
 			}
 		}
 		$crmtablealias = CRMEntity::getcrmEntityTableAlias($related_module);
-		$userNameSql = getSqlForNameInDisplayFormat(array('first_name'=> 'vtiger_users.first_name', 'last_name' => 'vtiger_users.last_name'), 'Users');
+		$userNameSql = getSqlForNameInDisplayFormat(array('enname'=> 'vtiger_users.ename'), 'Users');
```
denaldhushi, line 683: Fix the `enname` to `ename`

PR #1030, opened 2021-04-08
fix(Emails) relate mail to parent record when address is empty/erased…

george1806:
… and remove erased address from saved to


joebordes:
thanks

PR #1037, opened 2021-04-11
Get listview table ready for production

Luke1982:
I'll try to collect commits for the project in this PR, and keep every commit related to an issue (task) from the project.


joebordes:
a first review looks ok, except for the pagination.

have you tested ModTracker grid? to make sure the upgrade didn't break anything there? anywhere else we are using TUI that we should check?

Luke1982:
As far as I know these are the only two places we use TUI but good catch on modtracker, didn't test that. This PR is meant as an ongoing effort of the issues I'll pick up. The pagination will be, as discussed [here](https://github.com/tsolucio/corebos/issues/1038), some more work but we'll get that as well.

joebordes:
we are using it also extensively in the Business Question builder. There we are using the deprecated event also.

Luke1982:
OK I don't know about that module. Hopefully we can expect a repaired version soon

Luke1982:
Hey, that's cool. How did you hook into my PR?

denaldhushi:
@Luke1982 I cloned your repository and pushed the changes for collecting all new commits in this PR. that's all :)
Hope I don't have break any of your change ?

Luke1982:
Well if you cloned my repository I don't think you could've broken anything. But I am surprised that you were able to just push to my repository. I think I need to tweak some settings.

joebordes:
manually merged into master
thanks!

PR #1039, opened 2021-04-12
feat(cbMap) Generate DetailView Layout Map and DetailView Layout Developer Block

joebordes:


joebordes:
merged with all the other changes on your branch and updated with changes I made during validation

thanks!!

PR #1045, opened 2021-04-22
fix(Workflow) load folder names correctly in workflow engine

denaldhushi:


joebordes:
seems to be working ok. extra points for applying LDS formatting :clap: 

thanks!!

PR #1050, opened 2021-04-26
fix(Workflow) check for duplicate fields in update tasks

denaldhushi:


denaldhushi:
This PR is for update task workflow, in some cases when we have only one field, it shows that message and don't allow the save.

joebordes:
thanks!

PR #1058, opened 2021-05-04
fix(changeSets) Fixing Module List field length to module Global Vari…

shibe2018:
…able


joebordes:
discussed with Mohamed. Going to close this one in search of a more global solution.
Thanks

PR #1063, opened 2021-05-12
fix(Workflow) rename wrong function name

MiquelPerezGiner:


joebordes:
closing in favor of a8f38a931f852013019fc6d9150451bf9a2da99f which:

- doesn't break the [unit test](https://github.com/tsolucio/coreBOSTests/blob/master/modules/com_vtiger_workflow/expression_functions/stringsTest.php#L145)
- keeps the naming in line with the other new functions which also start with `cb_`

PR #1069, opened 2021-05-21
Export comments for specific user

shibe2018:


shibe2018:
for this to work we need to add excel export functionality that we already added on http://code.spike.studio:3000/LeroyMerlin/demo

PR #1073, opened 2021-05-28
fix(modules)field visibility in Layout Editor for blocks which are…

glorymoshi:
… not labeled


joebordes:
Can you please add the new label to the translation files?

joebordes:
thanks!

PR #1074, opened 2021-05-30
Feat/add inline on comments

geofreybundala:


modules/ModComments/ModCommentsCommon.js
```diff
@@ -66,9 +66,10 @@ if (typeof(ModCommentsCommon) == 'undefined') {
 				}
 			});
 		},
-		editCommentMode : function (parentId, commentId, comment) {
+		editCommentMode : function (parentId, commentId) {
 			document.getElementById('edit_comment_id_'+parentId).value = commentId;
-			document.getElementById('txtbox_'+parentId).value = comment;
+			var comment = document.getElementById('comment_content_'+commentId).innerHTML;
```
joebordes, line 71: do a direct assignment here, avoid storing the content in memory just to move it to the DOM, let JS do that internally if it needs to.

`document.getElementById('txtbox_'+parentId).value = trim(document.getElementById('comment_content_'+commentId).innerHTML);`

BTW, this commit and the code formatting one are very good!!

joebordes:
thanks!!

PR #1075, opened 2021-05-31
fix(cbMap) fix uitype 3314 field validation issue

shibe2018:


joebordes:
fixed with 4ca654d2dda88dd87de816c2639a58f8342bb112

PR #1093, opened 2021-06-24
Add support dollor sign filter on reports

shibe2018:


joebordes:
this change looks correct. I would like two things:

1.- add unit tests [here](https://github.com/tsolucio/coreBOSTests/blob/master/include/fields/DateTimeFieldTest.php) for the use case of `$` and `--$`

2.- looking at the change you made for MySQL 8 I ask myself: where else is that needed? Can you check if there are more date comparisons like that in the code and share what you find with me?  I am thinking mainly of Query Generator and Filters (customview) and Workflow Scheduler

PR #1122, opened 2021-08-31
feat(MasterDetail) check edit/delete permissions for records

denaldhushi:


joebordes:
This LGTM. I will accept it, but when Adisa informed me about this she said that even though the user did not have permission to delete the record, the record was deleted. That worried me as I can understand that we show a delete button/link in the frontend and make the call when we shouldn't but I would have expected the coreBOS backend to block and not permit the delete action.

Have you confirmed that a user with no permission to delete a record CAN delete it with a call from the frontend using the master-detail grid?

joebordes:
thanks!!

denaldhushi:
I can confirm that a user with no permission to delete can delete a record. I decided first to hide the buttons in DetailView if the user hasn't permissions. I will review that issue from backend why corebos don't stop the delete action and let you know.

PR #1139, opened 2021-09-27
Override lds modal cancel button

Luke1982:


Luke1982:
Goddammit it did it again

PR #1140, opened 2021-09-27
feat(ldsmodal) Allow overriding the text on the cancelbutton

Luke1982:


Luke1982:
Phew, OK after a hard reset and forced push to my repo, we're on the same path again. Must be more careful with my master branch

joebordes:
learning process...  so much fun :-)

LGTM
Thanks!

PR #1146, opened 2021-10-02
fix(ldsModal) Avoid redeclaring 'cancelButtonText'

Luke1982:
The previous commit led to a console warning and broke the ldsmodal


joebordes:
already had fixed this one. the CI called it out

Luke1982:
Very nice!

PR #1161, opened 2021-10-19
Retrying PDF Conversion when fails using GlobalVariable

tebajanga:


joebordes:
LGTM  Thanks!

#hacktoberfest-accepted

PR #1196, opened 2021-11-29
feat(Sendgrid) attach direct files in sendgrid integration

denaldhushi:


denaldhushi:
Send files via sendgrid for the "direct" option in send_mail method
https://github.com/tsolucio/corebos/blob/master/modules/Emails/mail.php#L25-L44

joebordes:
LGTM  Thanks!

PR #1203, opened 2021-12-09
Feat/support excel file type to bq t18004

george1806:


modules/cbQuestion/cbQuestion.php
```diff
@@ -849,5 +849,17 @@ public static function generateCSV($data, $delimiter = ',', $enclosure = '"') {
 		fclose($handle);
 		return $contents;
 	}
+
+	public static function getQnDelimeterProperty($qnid) {
+		$ans = self::getAnswer($qnid);
```
joebordes, line 854: This is very expensive in resources just to get the properties. We calculate the whole question's answer just to get the properties. Can we create another new static method in cbQuestion `getQuestionProperties`  that returns just the properties and then use that one in this method?

george1806: Sure, 'll update it.

PR #1226, opened 2022-01-14
feat(Pivot) add all available chart options

denaldhushi:


Smarty/templates/PivotView.tpl
```diff
@@ -11,7 +11,12 @@
 	{assign var='DESERTInfo' value='LBL_NO_DATA'|@getTranslatedString:$MODULE}
 	{include file='Components/Desert.tpl'}
 {else}
+<script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.5/d3.min.js"></script>
```
joebordes, line 14: Hi

We don't use external references like that. Download the code into the pivot directory with the others and load it from the local directory, please.

Thanks

joebordes:
Thanks!!

PR #1235, opened 2022-02-15
Ldsbuttons icons

egispike:


egispike:
Changed list view buttons with LDS and settings icons with LDS icons

PR #1238, opened 2022-02-18
feat(CustomView) add current user and group meta variable on advanced…

shibe2018:
… filter

### DIFF
--- a/modules/CustomView/CustomView.php
+++ b/modules/CustomView/CustomView.php
@@ -729,30 +729,24 @@ class CustomView extends CRMEntity {
 				) {
 					if ($advfilterval=='current_user') {
 						$advfilterval = trim($current_user->first_name.' '.$current_user->last_name);
+						$advfilterval = decode_html($advfilterval);
 					} else {
 						global $adb;
-						$result = $adb->pquery('select roleid from vtiger_users2group inner join vtiger_group2role on vtiger_users2group.groupid = vtiger_group2role.groupid where userid=?', array($current_user->id));
+						$advfilterval = array();
+						$result = $adb->pquery('select distinct roleid from vtiger_users2group inner join vtiger_group2role on vtiger_users2group.groupid = vtiger_group2role.groupid where userid=?', array($current_user->id));
 						$num_rows=$adb->num_rows($result);
 						for ($i=0; $i<$num_rows; $i++) {
 							$user_roleid=$adb->query_result($result, $i, 'roleid');
 						}
-						$result = $adb->pquery('select first_name, last_name from vtiger_users inner join vtiger_user2role on vtiger_users.id = vtiger_user2role.userid where roleid=?', array($user_roleid));
+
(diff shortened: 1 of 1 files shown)


modules/CustomView/CustomView.php
```diff
@@ -724,12 +725,34 @@ public function getAdvFilterByCvid($cvid) {
...
 				if (($col[1]=='smownerid' || $col[1]=='smcreatorid' || $col[1]=='modifiedby' || $uitype==101)
-					&& $advfilterval=='current_user' && (empty($_REQUEST['action']) || $_REQUEST['action']!='CustomView') && empty($_REQUEST['record'])
+					&& ($advfilterval=='current_user' || $advfilterval=='current_userandgroup') && (empty($_REQUEST['action']) || $_REQUEST['action']!='CustomView') && empty($_REQUEST['record'])
 				) {
-					$advfilterval = trim($current_user->first_name.' '.$current_user->last_name);
+					if ($advfilterval=='current_user') {
+						$advfilterval = trim($current_user->first_name.' '.$current_user->last_name);
+					} else {
+						global $adb;
+						$result = $adb->pquery('select roleid from vtiger_users2group inner join vtiger_group2role on vtiger_users2group.groupid = vtiger_group2role.groupid where userid=?', array($current_user->id));
```
joebordes, line 734: should we use `select distinct roleid from ....` here to reduce duplicate values?

modules/CustomView/CustomView.php
```diff
@@ -724,12 +725,34 @@ public function getAdvFilterByCvid($cvid) {
...
+						$num_rows=$adb->num_rows($result);
+						for ($i=0; $i<$num_rows; $i++) {
+							$user_roleid=$adb->query_result($result, $i, 'roleid');
+						}
+						$result = $adb->pquery('select first_name, last_name from vtiger_users inner join vtiger_user2role on vtiger_users.id = vtiger_user2role.userid where roleid=?', array($user_roleid));
```
joebordes, line 739: should we use `select distinct ename from ....` here to reduce duplicate values and extra work joining the name when we already have it in the ename column?

modules/CustomView/CustomView.php
```diff
@@ -724,12 +725,34 @@ public function getAdvFilterByCvid($cvid) {
...
+						$num_rows=$adb->num_rows($result);
+						for ($i=0; $i<$num_rows; $i++) {
+							$full_name[] = trim($adb->query_result($result, $i, 'first_name').' '.$adb->query_result($result, $i, 'last_name'));
```
joebordes, line 742: you should put here the decode_html operation to avoid having to do another loop below on line 749

modules/CustomView/CustomView.php
```diff
@@ -724,12 +725,34 @@ public function getAdvFilterByCvid($cvid) {
...
+						}
+					}
 					$advfilterval = decode_html($advfilterval);
 				}
-				$criteria['value'] = $advfilterval;
+				if (count($full_name) > 1) {
```
joebordes, line 747: it is faster to use empty instead of count

`if (!empty($full_name)) {`

PR #1265, opened 2022-04-13
feat(Fields) fix warning for uitype1025 and add pills component

denaldhushi:


joebordes:
Nice :-)

LGTM. Thanks!!

PR #1267, opened 2022-04-19
feat(Documents) relate documents with default folder if GV Document_DefaultFolder is set

denaldhushi:


modules/Documents/Documents.php
```diff
@@ -158,6 +158,15 @@ public function save_module($module) {
 		if (!empty($this->parentid)) {
 			$this->save_related_module('Documents', $this->id, getSalesEntityType($this->parentid), $this->parentid);
 		}
+		$Document_DefaultFolder = GlobalVariable::getVariable('Document_DefaultFolder', '');
+		if (empty($this->mode) && !empty($Document_DefaultFolder)) {
+			$folderId = __cb_getidof(array(
```
joebordes, line 163: I like the way you leverage the workflow system here :clap:

PR #1272, opened 2022-05-02
Rel module meta variable

george1806:


modules/com_vtiger_workflow/VTSimpleTemplate.inc
```diff
@@ -409,6 +411,15 @@ class VTSimpleTemplate {
 		$mod = CRMEntity::getInstance($modulename);
 		$wsid = $this->parent->getId();
 		list($wsid,$crmid) = explode('x', $wsid);
+		$maptype = '';
+		if (!empty($rlcondition)) {
+			$cbmap = cbMap::getMapByID($rlcondition);
+			if ($cbmap) {
+				$maptype = $cbmap->column_fields['maptype'];
+				if ($maptype == 'Condition Query') {
```
joebordes, line 419: @george1806 what is this empty "if" for?

joebordes: then delete it and add a comment like "we can add condition query support here in the future" and accept the PR

Thanks!!

PR #1292, opened 2022-05-25
fix(Webservices) special characters support

shibe2018:


joebordes:
closed in favor of translating correctly language files.

PR #1312, opened 2022-06-15
Fix (VTEmailTask) Using context variables for from email, from name, cc and bcc.

tebajanga:
…c and bcc


joebordes:
LGTM.  Thanks!

PR #1317, opened 2022-06-27
Add delivery time range msq

shibe2018:


include/cbmqtm/cbmqtm_dbdistributor.php
```diff
@@ -32,30 +32,34 @@ private static function setDB() {
 		static::$db = new PearDatabase();
 	}
 
-	public function sendMessage($channel, $producer, $consumer, $type, $share, $sequence, $expires, $deliverafter, $userid, $information) {
+	public function sendMessage($channel, $producer, $consumer, $type, $share, $sequence, $expires, $deliverafter, $userid, $information, $deliverRange=array('deliverStartTime'=>'', 'deliverEndTime'=>'', 'canSendOnSaturday'=>'', 'canSendOnSunday'=>'')) {
```
joebordes, line 35: this initialization is incorrect because if the programmer sends you a partial array:

`['canSendOnSaturday'=>0]`

which is valid, the other values are not initialized and will cause warnings and errors below. Please initialize the values individually.

include/cbmqtm/cbmqtm_dbdistributor.php
```diff
@@ -32,30 +32,34 @@ private static function setDB() {
...
 		}
 	}
 
-	private function insertMsg($channel, $producer, $consumer, $type, $share, $sequence, $expires, $deliverafter, $userid, $information) {
+	private function insertMsg($channel, $producer, $consumer, $type, $share, $sequence, $expires, $deliverafter, $userid, $information, $deliverRange) {
 		$rightnow = time();
 		if (empty($deliverafter)) {
 			$deliverafter = 0;
 		}
+		if (empty($deliverRange)) {
```
joebordes, line 55: if you get a partial array this "empty" will not apply but you won't get the other array elements initialized (as explained above).

Also, the initialization and the query below are wrong because you are forcing the deliver range start and end which is not the case. Most messages on the queue will not have this restriction and can be delivered at any hour of the day, any day. The start-end, and day conditions must apply only to those messages that require it.

I think you have to do the same "if" code you did for Saturday/Sunday for the start-end

include/cbmqtm/cbmqtm_dbdistributor.php
```diff
@@ -69,14 +73,18 @@ private function insertMsg($channel, $producer, $consumer, $type, $share, $seque
...
+				'canSendOnSaturday' => $deliverRange['canSendOnSaturday'],
+				'canSendOnSunday' => $deliverRange['canSendOnSunday'],
 		));
 	}
 
 	public function getMessage($channel, $consumer, $producer = '*', $userid = '*') {
 		self::setDB();
-		$sql = 'select * from cb_messagequeue where deliverafter<=? and channel=? and consumer=?';
-		$params = array(date('Y-m-d H:i:s', time()), $channel, $consumer);
+		$sql = 'select * from cb_messagequeue where deliverafter<=? and channel=? and consumer=? and deliverStartTime<? and deliverEndTime>?';
```
joebordes, line 86: The query below is wrong because you are forcing the deliver range start and end which is not the case. Most messages on the queue will not have this restriction and can be delivered at any hour of the day, any day. The start-end, and day conditions must apply only to those messages that require it.

I think you have to do the same "if" code you did for Saturday/Sunday for the start-end

include/cbmqtm/cbmqtm_dbdistributor.php
shibe2018: `
public function sendMessage($channel, $producer, $consumer, $type, $share, $sequence, $expires, $deliverafter, $userid, $information, $deliverRange = array()) {
`

include/cbmqtm/cbmqtm_dbdistributor.php
shibe2018: i added something like this 
`
$sqlcheck = 'select deliverStartTime, deliverEndTime from cb_messagequeue';
$resultcheck = static::$db->pquery($sqlcheck);
if ($resultcheck) {
	$sql .= ' and deliverStartTime<? and deliverEndTime>?';
}
`

PR #1321, opened 2022-07-05
feat(Dropzone) get title from the page url when we paste url in dropzone

denaldhushi:


denaldhushi:
I have added these parameters (120 seconds or we can change that to make it faster) to protect the code from the long time responding of any website.
```
CURLOPT_CONNECTTIMEOUT => 120,
CURLOPT_TIMEOUT => 120,
```
Normally after 120 seconds curl will response with an error code == 28 so we skip that process and save the url to notestitle.
I like that functionality to be always activated so we protect the notestitle from a LONG url, but for sure this is an optional choice and we can use a GV for that.

joebordes:
I hadn't noticed the timeouts. I am good with it and agree that we can add the parameter in the future as we get more feedback.

I am going to accept it.  thanks!!

PR #1324, opened 2022-07-13
Feat/security against xss

geofreybundala:


joebordes:
sorry, didn't see this PR in time. applied changes separately

PR #1334, opened 2022-07-27
Prevent PHPCS from showing an error on methods like 'vtlib_handler'

Luke1982:


joebordes:
Hi

I didn't add this exception because I actually want to enforce that naming convention for classes and functions. So all the new ones we create follow that guideline. Even if we can change some of the functions we have we align them with the camel case convention.

what do you think?

Luke1982:
I think following conventions is a good thing, so I agree. Do you know the comment you need to clear the error message for this specific message? Something like
```php
//phpcs:CamelCapsMethodName.NotCamelCaps
```?

**Comment by joebordes** [comment]
```diff
diff --git a/modules/Timecontrol/Timecontrol.php b/modules/Timecontrol/Timecontrol.php
index 70d32b8..137fc9e 100644
--- a/modules/Timecontrol/Timecontrol.php
+++ b/modules/Timecontrol/Timecontrol.php
@@ -203,7 +203,9 @@ class Timecontrol extends CRMEntity {
                }
        }
 
+       // phpcs:disable PSR1.Methods.CamelCapsMethodName.NotCamelCaps
        public function save_module($module) {
+       // phpcs:enable
                global $adb;
                if ($this->HasDirectImageField) {
                        $this->insertIntoAttachment($this->id, $module);
```

Luke1982:
This is easier:
```php
public function vtlib_handler($modulename, $event_type) { // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
```

PR #1338, opened 2022-08-04
Designing Vertical Menu

Klajd17:


Smarty/templates/Components/Components.tpl
```diff
@@ -206,3 +206,65 @@
 	{/foreach}
 	</ul>
 {/function}
+{literal}
```
joebordes, line 209: please move this javascript code to include/js/general.js so it gets cached and we can debug it

Smarty/templates/Components/Components.tpl
```diff
@@ -206,3 +206,65 @@
...
+function findUpModuleInMenu() {
+	let arr = {};
+	const modulename = document.querySelectorAll(`[data-name="${gVTModule}"]`);
+	if (modulename.length > 0) {
+		const parent = document.getElementById(`menu${modulename[0].dataset.level}`);
+		const childs = parent.querySelectorAll(`li span`);
+		modulename[0].parentElement.style.display = 'block';
+		modulename[0].style.background = '#c5d7e3';
+
+		//for (let i = 0; i < childs.length; i++) {
```
joebordes, line 228: please eliminate commented code

Smarty/templates/Header.tpl
```diff
@@ -212,25 +212,6 @@
 									</svg>
 									<span class="slds-assistive-text">{$APP.LNK_HELP}</span>
 							</button>
-							{if $SHOWQUICKCREATE}
```
joebordes, line 215: are you sure you have to delete this block?  I think this is the upper right corner quick create button and it is needed, or am I wrong?

Smarty/templates/Header.tpl
```diff
@@ -361,7 +342,6 @@
 	{$COREBOS_HEADER_PREMENU}
 	</div>
 	{/if}
-	{if $Application_Menu_Direction!='Vertical'}
```
joebordes, line 364: is this correct?  that change will make the code inside appear always

Smarty/templates/BrowserVariables.tpl
```diff
@@ -12,6 +12,7 @@
 var gVTModule = "{$GVTMODULE}";
 var gVTModuleLabel = "{$GVTMODULELABEL}";
 var gVTTheme  = "{$THEME}";
+var gVTModuleLabel = "{$GVTMODULELABEL}";
```
joebordes, line 15: next time make this commit and the change in include/utils/utils.php in a separate commit. we try to make small commits which are semantically related, so this one would have been

`feat(BrowserVariables) add module label`

you add that feature then you do another commit with the menu change that uses this feature

https://discussions.corebos.org/documentation/doku.php?id=en:devel:commitguidelines

include/js/general.js
```diff
@@ -91,6 +91,15 @@ GlobalVariable_getVariable('Application_Menu_Direction', 'Horizontal', (typeof g
 }, function (error) {
 	Application_Menu_Direction = 'Horizontal'; // set default value on error
 });
+
```
joebordes, line 94: this block is unnecessary, the code is already evaluating this global variable right above this block that you added and also the way you did it you have a different default value for the error.

on line 95 you set the default value to Vertical, then you call the getvariable with a default value of Vertical, but if that call fails, you set the default value to Horizontal on line 100. The code above sets the default value to Horizontal always which is how it should be.

include/js/general.js
```diff
@@ -7303,3 +7312,52 @@ $(function () {
 
 	new cbAccordion($('#cbmenu'), true);
 });
+
+
+function findParents(el) {
+	let parents = [];
+	while(el.parentNode) {
```
joebordes, line 7319: please [read our guidelines](https://discussions.corebos.org/documentation/doku.php?id=en:devel:developmentguidelines) on how to format the code

Smarty/templates/BrowserVariables.tpl
```diff
@@ -12,7 +12,7 @@
 var gVTModule = "{$GVTMODULE}";
 var gVTModuleLabel = "{$GVTMODULELABEL}";
 var gVTTheme  = "{$THEME}";
-var gVTModuleLabel = "{$GVTMODULELABEL}";
+var gVTModuleLabel = "{$GVTMODULELABEL}";
```
joebordes, line 15: This commit is incorrect. The extra space violates our guidelines. My comment was "next time". Next time separate the commit as it is a semantically separate change from the menu. Now it is to late to change the commit (or o costly to do so), so I was just trying to call your attention to being clean with your commits and changes, it will help you and us all going forward. Please revert this commit.

joebordes:
Thanks

PR #1347, opened 2022-08-25
fix(MailManager) add check request access for an authorized users

shibe2018:


joebordes:
looking for a more generic solution. thanks

PR #1354, opened 2022-09-21
feat(Events) Add type declaration to 'handleEvent' abstract method

Luke1982:


joebordes:
I don't want to do this yet. We still have a lot of installs on 7.2 and 7.3. It won't be long, but not yet

PR #1357, opened 2022-09-21
fix(CRMEntity) Make sure getInstance uses string modulename

Luke1982:


Luke1982:
After I updated my code, I could no longer save new assets due to the fact this method was called with an object when it was called for one of my custom modules. Now if there has been some change in the module structure, I think we should provide backwards compatibility or at least provide some deprecation warning.

joebordes:
the change you are referring to is probably this one in September 2020:

commit 31c48207afb3dda217fbdc6ae51cb70410e7f5a9
Author: joebordes <joe@tsolucio.com>
Date:   Wed Sep 9 02:26:55 2020 +0200

Where we eliminate an implicit object conversion to a string if it has a toString method. In my understanding, this is a hidden undocumented functionality that requires advanced object-oriented understanding to use. This is unnecessarily complex.

Your patch is similar, you are not making clear that the function accepts an object and you are forcing the object to have a `name` property. 

The method accepts a string as an input, not an object. Modify your code to use the method correctly.

Luke1982:
It's not my code. Some stock stuff is doing this. Here is the stack trace:
```
Fatal error: Uncaught Error: Object of class Vtiger_Module could not be converted to string in /public_html/crmgithubtest/data/CRMEntity.php:79 Stack trace:
  #0 /public_html/crmgithubtest/data/CRMEntity.php(3898): CRMEntity::getInstance()
  #1 /public_html/crmgithubtest/modules/com_vtiger_workflow/VTJsonCondition.inc(483): CRMEntity::getcrmEntityTableAlias()
  #2 /public_html/crmgithubtest/modules/com_vtiger_workflow/VTJsonCondition.inc(30): VTJsonCondition->checkCondition()
  #3 /public_html/crmgithubtest/modules/com_vtiger_workflow/VTWorkflow.php(210): VTJsonCondition->evaluate()
  #4 /public_html/crmgithubtest/modules/com_vtiger_workflow/VTEventHandler.inc(83): Workflow->evaluate()
  #5 /public_html/crmgithubtest/include/events/VTEventTrigger.inc(146): VTWorkflowEventHandler->handleEvent()
  #6 /public_html/crmgithubtest/includ in /public_html/crmgithubtest/data/CRMEntity.php on line 79
```

joebordes:
should be fixed with #1363

PR #1379, opened 2022-10-17
feat(BusinessActions) show custom alert message in Actions: workflow

denaldhushi:


Smarty/templates/DetailView.tpl
```diff
@@ -342,14 +344,14 @@ clipcopyobject.on('error', function(e) { clipcopyclicked = false; });
 																			<span class="slds-assistive-text">{$customlink_label}</span>
 																			</span>
 																		{else}
-																			<a class="webMnu" href="{$customlink_href}">
```
joebordes, line 345: should we urlencode or addslashes or something similar to avoid javascript errors and security injections?

build/changeSets/2022/addBAMessageFields.php
```diff
@@ -0,0 +1,50 @@
...
+			$this->sendMsg('Changeset ' . get_class($this) . ' already applied!');
+		} else {
+			$fields = array(
+				'BusinessActions' => array(
+					'LBL_BUSINESSACTIONS_SETTINGS' => array(
+						'error_msg' => array(
+							'columntype'=>'varchar(200)',
+							'typeofdata'=>'V~O',
+							'uitype'=>1,
+							'label' => 'Error Message',
```
joebordes, line 33: missing the translation of this label and the one below

Smarty/templates/DetailView.tpl
denaldhushi: I applied addslashes to avoid js errors. functions like urlencode or similiar we cannot apply here because in some use cases we are executing javascript functions in the actions, and these function will break the js function

PR #1389, opened 2022-10-20
refactor(App) replaced deprecated utf8_encode into mb_convert_encoding

maliknajjar:
## replaced deprecated utf8_encode into mb_convert_encoding

fun fact: we converted ISO-8859-1 into UTF-8 because These characters shown bellow (from 128 to 255) are stored using 1 byte for each character in ISO-8859-1. but in UTF-8 they are stored in multi byte space for each character.
That's why we need to convert it otherwise a UTF-8 encoder will not be able to read it.
ksnip_20221020-220336

### DIFF
--- a/include/js/general.js
+++ b/include/js/general.js
@@ -7290,10 +7290,10 @@ $(function () {
 		// Variables privadas
 		var links = this.el.find('.link');
 		// Evento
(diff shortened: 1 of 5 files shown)


joebordes:
LGTM.  Thanks!

PR #1391, opened 2022-10-23
feat(Documents) Adding push along flow buttons.

tebajanga:
For approving and declining the attached document on the record.


joebordes:
Thanks!!

PR #1399, opened 2022-10-28
security(VtlibUtils) clearing <a> tag href if URL is not valid

maliknajjar:
https://huntr.dev/bounties/f0272a31-9944-4545-8428-a26154d20348/


maliknajjar:
moved to VtlibUtils.php because I am not supposed to touch a package (HTMLPurifier.php)

joebordes:
https://github.com/ezyang/htmlpurifier/issues/369

joebordes:
this code is correct but I applied it manually because it was in the wrong place, the indentation was wrong and I had to collapse the two IFs. I added some unit tests too.

Thanks!!

PR #1403, opened 2022-11-04
fix(include) fix read only field save option

shibe2018:


include/js/FieldDependencies.js
```diff
@@ -550,6 +550,15 @@ FieldDependencies.prototype.fieldReadonlyDetailView = function (readonlyFields)
 		document.getElementById('dtlview_'+field).style.display='inline';
 		document.getElementsByName(field).item(0).setAttribute('readonly', true);
 		document.getElementsByName(field).item(0).style.display='inline';
+
+		let elem = document.getElementById('txtbox_'+field);
+		var siblings = [];
+		while (elem = elem.nextElementSibling) {
```
joebordes, line 556: Why two loops?  Can't you just directly set the display in the first loop? as far as I see you don't use that siblings array for anything else.

Don't you have to "undo" this?  I mean, this is when a set of fields have to be hidden on a dependency, I understand that if that dependency value changes you would have to set these fields back to display=block, no?

shibe2018: Changed to this
```js
let elem = document.getElementById('txtbox_'+field);
	while (elem = elem.nextElementSibling) {
	elem.style.display = 'none';
}
```
This for hide `save` and `cancel` buttons when fields are `readonly` on dependency

shibe2018: and trying to edit on that field

shibe2018:
sure your right let me do that thank you

PR #1413, opened 2022-11-11
fix(AdvancedSearch) sending map information in the Popup.php as well

maliknajjar:
Fix for the Advanced Search in Popup View


joebordes:
Thanks!

PR #1416, opened 2022-11-14
feat(Mobile) Access both edit and detail view from listView

maliknajjar:
In Mobile App:
* When clicking on the records, direct us to the detail view.
* When we click the arrow at the end of the row, it directs us to the edit view.

image


modules/Mobile/actions/getScrollContent.php
```diff
@@ -62,8 +62,16 @@ public function process(crmtogo_API_Request $request) {
 			}
 			$id = $ws_entity2.'x'.$adb->query_result($records, $i, $entityidfield);
 
-			$output .= '<li><a class="ui-btn ui-btn-icon-right ui-icon-carat-r" target="_self" href="?_operation=fetchRecord&record='.$id.'" target="_self">'
-				.$firstname.'&nbsp;'.$lastname.'</a></li>';
+			$fetchRecordUrl = '/modules/Mobile/index.php?_operation=fetchRecord&record=' . $id;
```
joebordes, line 65: Are you sure that the absolute path is correct in all cases? Can we leave it the way it was, just the query parameters?

`$fetchRecordUrl = '?_operation=fetchRecord&record=' . $id;`

instead of

`$fetchRecordUrl = '/modules/Mobile/index.php?_operation=fetchRecord&record=' . $id;`

just to make it as similar to the way it was as possible.

maliknajjar: ok I will change it

joebordes:
Thanks!!

PR #1417, opened 2022-11-14
fix(AdvancedFilter) setting advancedSearchData value to 'MAP_NOT_FOUND' if null

maliknajjar:
This pull request fixes Advanced search in create filter page.


joebordes:
Thanks!

PR #1419, opened 2022-11-16
feat(Popup) filter records in popup between 2 modules

denaldhushi:
This is an enhance of Popup query hook (previously done for fields), now we can use conditions directly in a ListColumns map to filter records between 2 modules in a related list popup view. This map is an example of relation of Accounts with Documents, and we filter records of Documents only when we open the Popup in Accounts module.
* Map (conditions for related modules in related lists view):
* MapName: `Accounts_ListColumns`
```xml
<map>
	<originmodule>
		<originname>Accounts</originname>
	</originmodule>
	<popup>
		<conditions>
			<condition>
				<formodule>Accounts</formodule>
				<relatedmodule>Documents</relatedmodule>
				<value>[{"fieldname":"filename","operation":"contains","value":"t","valuetype":"rawtext","joincondition":"or","groupid":"0"}]</value>
			</condition>
		</conditions>
	</popup>
</map>
```

### DIFF
--- a/data/CRMEntity.php
+++ b/data/CRMEntity.php
@@ -1646,14 +1646,14 @@ class CRMEntity {
 			if (!empty($conditions[$fieldname]) && json_decode($conditions[$fieldname]) == null) {
 				return $conditions[$fieldname];
 			}
-			if (!empty($conditions[$fieldname])) {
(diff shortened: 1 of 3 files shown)


joebordes:
Nice :-D

Thanks!

PR #1420, opened 2022-11-17
feat(cbMaps) PopupFilter BusinessMap

maliknajjar:
## PopupFilter BusinessMap
Create a business map that adds some special filtering or conditions on the records that will appear in the popup screen.
the structure of the map would be similar to this:
```xml
<modulename> (the related module that we're filtering on)
 <fieldname> (the field that we click on)
  <conditions> (conditions that are going to be applied on)
   <condition>
```
The name of the map will be **modulename_PopupFilter**

### DIFF
--- a/Smarty/templates/DetailView.tpl
+++ b/Smarty/templates/DetailView.tpl
@@ -28,6 +28,11 @@ clipcopyobject.on('success', function(e) { clipcopyclicked = false; });
 clipcopyobject.on('error', function(e) { clipcopyclicked = false; });
 {/literal}
 </script>
+<script>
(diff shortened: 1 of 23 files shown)


Smarty/templates/DetailView.tpl
```diff
@@ -28,6 +28,11 @@ clipcopyobject.on('success', function(e) { clipcopyclicked = false; });
 clipcopyobject.on('error', function(e) { clipcopyclicked = false; });
 {/literal}
 </script>
+<script>
```
joebordes, line 31: move script tags inside IF

joebordes:
Thanks!!

PR #1421, opened 2022-11-17
fix(advanceFilter) show saved conditions in advanceFilter

maliknajjar:
sometimes the "input.value" is an array and sometimes it is an object


joebordes:
Thanks

PR #1422, opened 2022-11-18
i18n(Documents) translating labels of PDF previewer

shibe2018:


joebordes:
please add these strings to all the other translation files also. Copy them in English, the translators will take care of each specific language

PR #1423, opened 2022-11-22
Fix uitype10 trigger change issue

shibe2018:


joebordes:
it isn't equivalent. I mean, in detail view we clear by setting a '0', while now you are setting an empty string. I understand that it will work as we will save an "empty" value in the database but I would like to ensure you have tried it, right? You have inline edited a uitype 10 field, cleared it, saved it, and reloaded it to make sure it is now empty. Hopefully also looked in the database to make sure we are saving the value consistently.  Right?

joebordes:
thanks

shibe2018:
sure

PR #1426, opened 2022-11-23
feat(AdvancedSearch) ability to add fields from other modules

maliknajjar:
adding the ability to specify fields from other modules in the advancedSearch cbMap XML


joebordes:
Thanks!

PR #1432, opened 2022-11-28
feat(GLobalVariables) adding the Application_ListView_FilterPanel_Open

maliknajjar:
We need to create a global variable (Application_ListView_FilterPanel_Open) that takes [ 0 | 1 ] as values.
When the GV is set to 0, it should:

* hide thee filter's panel as shown in the image below:

### DIFF
--- a/modules/Vtiger/ListView.php
+++ b/modules/Vtiger/ListView.php
@@ -276,6 +276,8 @@ $smarty->assign('advancedSearchMapResult', $advancedSearchMapResult);
 $Application_ListView_FilterPanel_Open = GlobalVariable::getVariable('Application_ListView_FilterPanel_Open', '1', $currentModule, '', $_REQUEST['action']);
 if (!$Application_ListView_FilterPanel_Open) {
 	$smarty->assign('Application_ListView_FilterPanel_Open', "display: none;");
+} else {
+	$smarty->assign('Application_ListView_FilterPanel_Open', "");
 }
 
 if (!empty($custom_list_include) && file_exists($custom_list_include)) {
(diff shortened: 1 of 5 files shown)


modules/Vtiger/ListView.php
```diff
@@ -272,6 +272,12 @@
 }
 $smarty->assign('advancedSearchMapResult', $advancedSearchMapResult);
 
+// GV responsible for showing or hiding filter panel
+$Application_ListView_FilterPanel_Open = GlobalVariable::getVariable('Application_ListView_FilterPanel_Open', '1', $currentModule, '', $_REQUEST['action']);
+if (!$Application_ListView_FilterPanel_Open) {
+	$smarty->assign('Application_ListView_FilterPanel_Open', "display: none;");
+}
```
joebordes, line 279: you need an "else" here to set the Smarty variable in the case it is not open

modules/Vtiger/ListView.php
```diff
@@ -272,6 +272,12 @@
 }
 $smarty->assign('advancedSearchMapResult', $advancedSearchMapResult);
 
+// GV responsible for showing or hiding filter panel
+$Application_ListView_FilterPanel_Open = GlobalVariable::getVariable('Application_ListView_FilterPanel_Open', '1', $currentModule, '', $_REQUEST['action']);
```
joebordes, line 276: I think the logic is inverted here
the default value is "1! but in that case you set the smarty variable to "display none"  
Is that correct?

maliknajjar: I set it to display none if it is not "1" `!$Application_ListView_FilterPanel_Open`

modules/Vtiger/ListView.php
maliknajjar: the reason why I kept it like that is because Smarty did not crash .. but I will change it right away

joebordes:
thanks!!

PR #1434, opened 2022-11-29
fix(ListView) resolve error when recordId is not given in BusinessActions

maliknajjar:
BA issue in ListView....when the BA calls a business question it needs a specific record id to work.. here is an image if we don't set a specific record id, the list view doesn't get loaded: image the business question: https://ncarefit.evolutivo.it/index.php?module=cbQuestion&action=DetailView&record=8616 the business action: block://bqanswer:modules/cbQuestion/cbQuestionWidget.php:QID=8616&RECORDID=8551

### DIFF
--- a/modules/cbQuestion/cbQuestionWidget.php
+++ b/modules/cbQuestion/cbQuestionWidget.php
@@ -43,7 +43,7 @@ class bqAnswer_DetailViewBlock extends DeveloperBlock {
 				'$MODULE$' => $module,
 				'$USERID$' => $current_user->id,
 			);
-			if (isset($recordid) && $recordid != '0x0') {
+			if (isset($recordid) && $recordid != '0x0' && !empty($recordid)) {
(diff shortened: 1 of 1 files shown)


modules/cbQuestion/cbQuestionWidget.php
```diff
@@ -43,7 +43,7 @@ public function process($context = false) {
 				'$MODULE$' => $module,
 				'$USERID$' => $current_user->id,
 			);
-			if (isset($recordid) && $recordid != '0x0') {
+			if (isset($recordid) && $recordid != '0x0' && !empty($recordid)) {
```
joebordes, line 46: should be

`if (!empty($recordid) && $recordid!='0x0') {`

because if the variable is !empty include isset  (I think)

maliknajjar: I think isset checks if it is declared .. not if it is empty
image

joebordes:
Thanks!!

PR #1435, opened 2022-11-30
fix(PopupFilter) deep cloning the arr object

maliknajjar:
The nature of javascript is to treat variables that are objects and arrays as refrences (pointers) when passed as a function parameter. That led to mutating the original array. which had unexpected behavior. The issue was fixed by deeply cloning the array using the structuredClone() function. This function clones the array itself and all the objects inside of it. Because cloning the array alone and keeping the same value (in this case they are objects' refrences) inside of it is not sufficient to prevent mutating the original variable.


joebordes:
Thanks!!

PR #1440, opened 2022-12-13
feat(DetailView) sort widgets and blocks by sequence, apply some LDS

denaldhushi:
dv3 png
dv1
dv2


joebordes:
wow!  this is a great piece of work. humbled.  thank you!!

PR #1443, opened 2022-12-15
feat(PopupFilter) get value from inside uitype10 fields

maliknajjar:
This PR gives the ability to take values from inside a uitype10 record.
you can get inside any amount of uitype10 fields to get a wanted field value.

### example
```
$account_id->$account_id->$phone
```

### DIFF
--- a/modules/Vtiger/ExecuteFunctions.php
+++ b/modules/Vtiger/ExecuteFunctions.php
@@ -693,6 +693,31 @@ switch ($functiontocall) {
 			);
 		}
 		break;
+	case 'getDeepFieldValue':
+		$moduleName = vtlib_purify($_REQUEST['moduleName']);
+		$value = vtlib_purify($_REQUEST['value']);
+		$fieldsArray = explode("-&gt;", $value);
+		// remove all $ signs
+		foreach ($fieldsArray as &$value) {
+			$value = substr($value, 1);
+		}
+		$firstFieldRecordID = vtlib_purify($_REQUEST['firstFieldRecordID']);
+
+		$result = '';
+		$currentFieldvalue = $firstFieldRecordID;
+		foreach ($fieldsArray as $key => $fieldName) {
+			if ($key == 0) {
+				continue;
+			}
+			$queryGenerator = new QueryGenerator(getSalesEntityType($currentFieldvalue), $current_user);
+			$queryGenerator->setFields(array($fieldName));
+			$queryGenerator->addCondition('id', $currentFieldvalue, 'e');
+			$query = $queryGenerator->getQuery();
+			$result = $adb->query($query)->fields[0];
+			$currentFieldvalue = $result;
+		}
+		$ret = $result;
+		break;
 	default:
 		$ret = '';
 		break;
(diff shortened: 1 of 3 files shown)


modules/cbMap/processmap/PopupFilter.php
```diff
@@ -62,6 +62,19 @@
...
+
+	/////////////////////////////////////////////////////////////
+	// example for getting values from inside uitype10 fields: //
+	/////////////////////////////////////////////////////////////
+	<map>
+		<modulename>Contacts</modulename>
+		<field>
+			<fieldname>account_id</fieldname>
+			<modulename>Accounts</modulename>
+			<advft_criteria>[{"groupid":1,"columnname":"contact","comparator":"e","value":"$account_id->$account_id->$phone","columncondition":""}]</advft_criteria>
```
joebordes, line 74: can we use `Accounts.phone` like we do in other parts of the application?

modules/Vtiger/ExecuteFunctions.php
```diff
@@ -693,6 +693,31 @@
...
+		}
+		$firstFieldRecordID = vtlib_purify($_REQUEST['firstFieldRecordID']);
+
+		$result = '';
+		$currentFieldvalue = $firstFieldRecordID;
+		foreach ($fieldsArray as $key => $fieldName) {
+			if ($key == 0) {
+				continue;
+			}
+			$queryGenerator = new QueryGenerator(getSalesEntityType($currentFieldvalue), $current_user);
```
joebordes, line 712: this is expensive, calling the class and launching a query inside the loop for each field when we can do one query  and the whole function overlaps a lot with `getFieldValuesFromRecord`, look for me and we discuss

modules/Vtiger/ExecuteFunctions.php
```diff
@@ -693,6 +693,31 @@
 			);
 		}
 		break;
+	case 'getDeepFieldValue':
+		$moduleName = vtlib_purify($_REQUEST['moduleName']);
+		$value = vtlib_purify($_REQUEST['value']);
+		$fieldsArray = explode("-&gt;", $value);
+		// remove all $ signs
+		foreach ($fieldsArray as &$value) {
```
joebordes, line 701: if substitute this loop with array_walk, would it be faster?

`array_walk($fieldsArray, function (&$item, $key) { $item = substr($item, 1); });`

modules/cbMap/processmap/PopupFilter.php
maliknajjar: ok

maliknajjar:
I added the ability to select multiple fields. for the popupFilter usecase we dont need it. but maybe we will need it for other usecases in the future.
example: `$account_id.$account_id.$phone,account_id`

joebordes:
this is a really interesting piece of work. the abstraction you do there is a clear sign that you get what coreBOS is trying to do. fantastic!

I also like the idea of separating the record relation levels with dots going down as needed. We could try to apply the same idea in other parts of the application.

Thanks!

PR #1445, opened 2022-12-19
fix(PopupFilter) make popup_filter_map_popup_window return value not promise

maliknajjar:
async functions always return a promise. so the popup_filter_map_popup_window does not return the boolean values unless ***await*** is used with it. the solution is to remove the async keyword from the function and to return the boolean value immediately once the condition is met. after the boolean value is returned the promise will be executed and it will fetch the results from the backend and then open the popup.


joebordes:
LGTM
Thanks!

PR #1448, opened 2022-12-22
Watermarktask

maliknajjar:
WF to add watermarks on images in corebos.
you can set the size of the watermark image. it is a percentage value based on the size of the main image.
for the position field you have these options: `center, top, bottom, right, left, topright, topleft, bottomleft, bottomright`

### DIFF
--- a/modules/com_vtiger_workflow/language/en_us.lang.php
+++ b/modules/com_vtiger_workflow/language/en_us.lang.php
@@ -224,6 +224,12 @@ $mod_strings = array(
 'HELP_GenerateImageCode_ValueToEncode' => 'Field name where the value to encode will be read from',
 'HELP_GenerateImageCode_EncodingType' => 'Encoding type to be used for the image',
 'HELP_GenerateImageCode_SaveTo' => 'Image field where the image will be saved',
+// watermark workflow context variable
+'HELP_WaterMark_Image_value' => 'Watermark image url',
+'HELP_Image_Field_Name' => 'The field of the image that gets watermarked',
+'HELP_WaterMark_Size' => 'The size of the watermark image in percentage',
+'HELP_WaterMark_Position' => 'the position of the watermark in image',
+'HELP_Expression_Type' => 'expresstion type',
 'CONTEXT_VARIABLES' => 'Context variables',
 'LBL_ON_RELATE' => 'On Relate with',
 'LBL_ON_UNRELATE' => 'On Unrelate with',
--- a/modules/com_vtiger_workflow/language/en_us.lang.php
+++ b/modules/com_vtiger_workflow/language/en_us.lang.php
@@ -225,6 +225,9 @@ $mod_strings = array(
 'HELP_GenerateImageCode_EncodingType' => 'Encoding type to be used for the image',
 'HELP_GenerateImageCode_SaveTo' => 'Image field where the image will be saved',
 // watermark workflow context variable
+'Watermark' => 'Watermark',
+'CBWatermark' => 'Watermark',
+'HELP_CBWatermark' => 'This workflow task will apply a watermark to the image contained in the given field',
 'HELP_WaterMark_Image_value' => 'Watermark image url',
 'HELP_Image_Field_Name' => 'The field of the image that gets watermarked',
 'HELP_WaterMark_Size' => 'The size of the watermark image in percentage',
--- a/modules/com_vtiger_workflow/tasks/CBWatermark.php
+++ b/modules/com_vtiger_workflow/tasks/CBWatermark.php
@@ -24,31 +24,47 @@ class CBWatermark extends VTTask {
 	public $queable = true;
 
 	public function getFieldNames() {
-		return array('imagesvalue', 'imagesx', 'imagesy', 'exptype');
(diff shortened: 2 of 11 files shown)


modules/com_vtiger_workflow/tasks/CBWatermark.php
```diff
@@ -0,0 +1,172 @@
...
+			$waterMarkDotArr = explode('.', $waterMarkUrl);
+			$waterMarkName = $waterMarkArr[count($waterMarkArr) - 1];
+			$waterMarkType = $waterMarkDotArr[count($waterMarkDotArr) - 1];
+			$waterMarkSavedPath = $root_directory . 'storage/';
+			$data = $entity->getData();
+			$mainImageFileName = $data['filename'];
+			$mainImageFileType = $data['filetype'];
+			$mainImageFileSize = $data['filesize'];
+			$mainImageDownloadUrl = $data['_downloadurl'];
+			$mainImagePath = substr($root_directory, 0, strlen($root_directory) - 9) . parse_url($mainImageDownloadUrl)['path'];
```
joebordes, line 63: if I understand correctly these last 5 lines, this task only works with the Documents module, not any image field on any module. If this is the case, we should permit this workflow only for the Documents module adding the module to the changeset

https://github.com/tsolucio/corebos/pull/1448/files#diff-c342bcb94dfa45251e6927c5ecebae0927c68861ff30ca9f6de051dd2f2e4eabR28

or modify the task to permit selecting the image field we want to apply the watermark to

joebordes:
I understand that we overwrite the image in the disk. That means we lose the original image. Since Documents only has one image field this could be a problem. In order to have both images we need to add some fields to permit setting conditions on which Document records to apply the watermark to. That makes sense but we have to document that and that this task only works with Documents.

joebordes:
- missing translations, something [like these:](https://github.com/tsolucio/corebos/blob/master/modules/com_vtiger_workflow/language/en_us.lang.php#L220-L223)
- missing some more log messages
- missing [context variables](https://github.com/tsolucio/corebos/blob/master/modules/com_vtiger_workflow/tasks/GenerateImageCodeWorkflowTask.inc#L37) ??

modules/com_vtiger_workflow/tasks/CBWatermark.php
```diff
@@ -0,0 +1,311 @@
...
+			// 	$expression = $parser->expression();
+			// 	$exprEvaluater = new VTFieldExpressionEvaluater($expression);
+			// 	$watermark = $exprEvaluater->evaluate($entity);
+			// }
+			$waterMarkUrl = $watermark;
+			$waterMarkArr = explode('/', $waterMarkUrl);
+			$waterMarkDotArr = explode('.', $waterMarkUrl);
+			$waterMarkName = $waterMarkArr[count($waterMarkArr) - 1];
+			$waterMarkType = $waterMarkDotArr[count($waterMarkDotArr) - 1];
+			$waterMarkSavedPath = $root_directory . 'storage/';
```
joebordes, line 137: this hard-coded `storage` should be:  `GlobalVariable::getVariable('Application_Storage_Directory', 'storage/')`

modules/com_vtiger_workflow/language/en_us.lang.php
```diff
@@ -224,6 +224,12 @@
 'HELP_GenerateImageCode_ValueToEncode' => 'Field name where the value to encode will be read from',
 'HELP_GenerateImageCode_EncodingType' => 'Encoding type to be used for the image',
 'HELP_GenerateImageCode_SaveTo' => 'Image field where the image will be saved',
+// watermark workflow context variable
+'HELP_WaterMark_Image_value' => 'Watermark image url',
+'HELP_Image_Field_Name' => 'The field of the image that gets watermarked',
+'HELP_WaterMark_Size' => 'The size of the watermark image in percentage',
+'HELP_WaterMark_Position' => 'the position of the watermark in image',
+'HELP_Expression_Type' => 'expresstion type',
```
joebordes, line 232: we have to add here the name, label, and class of the workflow task, they are used in some parts of the workflow system

```php
'Watermark' => 'Watermark',
'CBWatermark' => 'Watermark',
```

and also

```php
'HELP_CBWatermark' => 'This workflow task will apply a watermark to the image contained in the given field',
```

 (or something like that). This one is used in the Help screen.

modules/com_vtiger_workflow/tasks/CBWatermark.php
```diff
@@ -0,0 +1,311 @@
...
+		$currentRecordId = explode('x', $entity->id)[1];
+		$from_wf = true;
+		$hold_ajxaction = isset($_REQUEST['ajxaction']) ? $_REQUEST['ajxaction'] : '';
+		$_REQUEST['ajxaction'] = 'Workflow';
+		if (!empty($wmImageValue)) {
+			$watermark = '';
+			if ($exptype == 'rawtext') {
+				$watermark = $wmImageValue;
+			}
+			// elseif ($exptype == 'fieldname') {
```
joebordes, line 120: please eliminate the commented code and the `if ($exptype == 'rawtext') {`  above.

joebordes:
thanks!

PR #1450, opened 2023-01-04
feat(Pivot) make rows header in pivot clickable

denaldhushi:


include/utils/utils.php
```diff
@@ -3904,4 +3904,20 @@ function validateClickHouseSecret($signedvalue, $signedkey, $input) {
 	$computedSignature = base64_encode(hash_hmac('sha256', $input, $signedvalue, true));
 	return ($receivedSignature === $computedSignature);
 }
+
+function getFieldForAdvancedSearch($field, $module) {
```
joebordes, line 3908: we have this function in Custom View: https://github.com/tsolucio/corebos/blob/master/modules/CustomView/CustomView.php#L530

please use that one

Smarty/templates/PivotView.tpl
```diff
@@ -121,6 +122,24 @@ $(function() {
...
+					let vColRowLabels = vTotalAxisLabels - vTotalRowLabels ;
+					let thHeaderName = $('.pvtAxisLabel').eq($(this).index()+vColRowLabels);
+					const advSearch = [{
+						'columnname': rowFields[thHeaderName.html()],
+						'comparator': 'e',
+						'value': e.target.innerText,
+						'groupid': 1,
+						'columncondition': 'and'
+					}];
+					const urlstring = `${JSON.stringify(advSearch)}&advft_criteria_groups=[null,{"groupcondition":""}]`;
```
joebordes, line 140: don't we have to use encodeuricomponent here?  For example, if `e.target.innerText` or the header has a value of `me&you` I think that ampersand will break the URL.

denaldhushi: Updated :D

joebordes:
LGTM. Thanks!

PR #1451, opened 2023-01-04
feat(GlobalVariable) adding Application_Menu_Show

maliknajjar:
Create the Application_Menu_Show GV
When the GV is set to 0, the menu should not be visible


joebordes:
thanks!!

PR #1452, opened 2023-01-04
feat(GLobalVariable) adding Application_Toolbar_Show

maliknajjar:
Create the GV Application_Toolbar_Show
When the GV is set to 0, the buttons in the red rectangle should not be shown


joebordes:
Thanks!

PR #1453, opened 2023-01-05
feat(WorkFlow) show available image fields in picklist

maliknajjar:
show available image fields in picklist
image


modules/Vtiger/ExecuteFunctions.php
```diff
@@ -693,6 +693,12 @@
 			);
 		}
 		break;
+	case 'getModuleFieldsInfo':
```
joebordes, line 696: I don't want to expose this functionality to JS. They can use web service to get that information. Leave just the change above please

maliknajjar:
image

joebordes:
thanks!

PR #1454, opened 2023-01-06
feat(Emails) Add aftersend action to PHPMailer e-mail sending

Luke1982:


modules/Emails/Emails.php
```diff
@@ -298,6 +298,9 @@ public static function sendEMail(
 			$mail->AddReplyTo($HELPDESK_SUPPORT_EMAIL_REPLY_ID);
 		}
 		$mail_status = MailSend($mail);
+
+		cbEventHandler::do_action('corebos.email.aftersend', $mail);
```
joebordes, line 302: add the $mail_status to the parameters that the event receives and I will accept. LGTM

Luke1982: Can we do that by just adding a parameter to the `do_action` call or do I need to wrap it in an array?

joebordes: has to be wrapped in an array

Luke1982: Done. The receiving function should unwrap that array right? That doesn't happen automagically?

joebordes:
Thanks!!

modules/Emails/Emails.php
joebordes: Yes, correct

PR #1456, opened 2023-01-09
feat(Pivot) back to last clicked url

denaldhushi:


joebordes:
LGTM and I will accept but it catches my attention that you are not using the `backtourl`  function you create.  Is that controlled?

denaldhushi:
`backtourl` is created as a BusinessAction. But for sure we can use that function in the future if we need it.

PR #1458, opened 2023-01-11
feat(ListView) Added GV for toggling the Mass Delete/Edit buttons

asubaruwrxsti:


joebordes:
Hi

Welcome to the coreBOS contributors !! :-D

This PR LGTM, I have some comments

- Please add the GV definition also to the fr_fr.gvdefs.php file (in the same place)
- I think you need to add the GV Smarty assign also to [modules/Documents/ToastListView.php](https://github.com/tsolucio/corebos/blob/master/modules/Documents/ToastListView.php#L161)
- Call your attention that the permission of these two buttons is set in the [getListButtons](https://github.com/tsolucio/corebos/blob/master/data/CRMEntity.php#L3760) method. you could have solved the request by modifying that method. Adding the GV call there. I like the solution you have made better, so I'm not asking you to change anything, only sharing another possible approach so you can learn a little more about the application.

Thanks!

joebordes:
Thanks!!

PR #1459, opened 2023-01-12
feat(GlobalVariables) Adding Application_DetailView_ActionPanel_CenteredActions GV

asubaruwrxsti:


joebordes:
I do not approve this type of change in the base product. This is not what global variables are for. In my mind, this is a customization for your particular use case, not something everybody could want nor either something that I want to maintain for the rest of the life of the project. if we decide to make all the buttons the same size, following LDS guidelines we do it for all, but not like this.

have you tried with widgets? have you tried with links? how does this CSS affect those?

we can discuss this a bit more if you want. Look for me.

PR #1461, opened 2023-01-13
fix(PopupFilter) force Popup Window to always open a new window

maliknajjar:
if you try to open a popup window from within a popup window with the same name ('vtlibui10qc') the "popup_filter_map_popup_window" will redirect you to the url instead of openning a new window.


joebordes:
Thanks!!

PR #1468, opened 2023-01-19
fix(TUI Grid) for RelatedList

ochu-elhadji:
- Displaying full text in tui-grid-cell for related list
- Fixing tui-grid-cell on editing mode
image


joebordes:
Thanks!

PR #1472, opened 2023-01-20
fix(Tui-grid-table) table header overlapping when columns names are t…

ochu-elhadji:
Fixing the column header:
image


joebordes:
thanks!

PR #1475, opened 2023-01-20
refactor(CoreBOS) change deprecated PHP function utf8_decode

maliknajjar:
change deprecated PHP function utf8_decode


joebordes:
thanks!

PR #1477, opened 2023-01-22
feat(expression) support expression inside strings

maliknajjar:
## Support expression inside strings

You can use curly brackets  `{ expression }` to write valid expressions inside strings. This feature made for long string parameters like the condition parameter of the aggregation function. That parameter supports selecting uitype10 fields but it does not support the usage of normal expression functions.
Inspired by javascript [Template Literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

### Features
* Works inside strings
* Supports nesting
* Works outside strings too (similar to a Preprocessor in c/c++)

### Examples
image
image
image
image

### DIFF
--- a/modules/com_vtiger_workflow/VTWorkflowUtils.php
+++ b/modules/com_vtiger_workflow/VTWorkflowUtils.php
@@ -59,7 +59,7 @@ class VTWorkflowUtils {
 			if (trim($fieldValue)=='') {
 				$fieldValue = '';
 			} else {
-				$parser = new VTExpressionParser(new VTExpressionSpaceFilter(new VTExpressionTokenizer($fieldValue)));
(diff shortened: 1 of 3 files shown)


joebordes:
This is awesome!!  Not only the idea which makes the expression a lot easier but the simplicity of the implementation. Great job!

I have one issue which is the election of symbols. The opening and closing keys are relatively standard and can conflict in some workflows, especially for web services workflow tasks, but also we use these in GenDoc and they will not work correctly in the {expression} directive (an edge case but valid). Can we change them to something else? backticks or some others?

maliknajjar:
what about normal brackets `[ expression ]`or two curley brackets `{{ expression }}`? 
the problem with backticks `` `expression` `` is that its hard to notice or read sometimes.

maliknajjar:
or we can start it with a dollar sign `${ expression }` or a hashtag `#{ expression }`

joebordes:
Thanks!

PR #1480, opened 2023-01-24
fix(tuiGrid) changing condition priorities to fix manual sorting

maliknajjar:
changing condition priorities to fix manual sorting
image


joebordes:
Thanks!

PR #1482, opened 2023-01-26
feat(GlobalVariables) enable or disable creating a new document in storage and Documents module

maliknajjar:
enable or disable creating a new document in storage and Documents module

### DIFF
--- a/include/utils/utils.php
+++ b/include/utils/utils.php
@@ -3905,11 +3905,13 @@ function validateClickHouseSecret($signedvalue, $signedkey, $input) {
 	return ($receivedSignature === $computedSignature);
 }
 
-function convertPathFromAbsoluteToReltive($absoltePath) {
+function convertPathFromAbsoluteToRelative($absolutePath) {
 	global $root_directory;
-	$prefix_pos = strpos($absoltePath, $root_directory);
-	$rel_path = substr($absoltePath, $prefix_pos + strlen($root_directory));
-	return $rel_path;
+	$prefix_pos = strpos($absolutePath, $root_directory);
+	if (!$prefix_pos) {
+		return "";
+	}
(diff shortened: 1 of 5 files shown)


include/utils/utils.php
```diff
@@ -3904,4 +3904,12 @@ function validateClickHouseSecret($signedvalue, $signedkey, $input) {
 	$computedSignature = base64_encode(hash_hmac('sha256', $input, $signedvalue, true));
 	return ($receivedSignature === $computedSignature);
 }
+
+function convertPathFromAbsoluteToReltive($absoltePath) {
```
joebordes, line 3908: Relative

include/utils/utils.php
```diff
@@ -3904,4 +3904,12 @@ function validateClickHouseSecret($signedvalue, $signedkey, $input) {
...
+	global $root_directory;
+	$prefix_pos = strpos($absoltePath, $root_directory);
```
joebordes, line 3910: absolute

include/utils/utils.php
```diff
@@ -3904,4 +3904,12 @@ function validateClickHouseSecret($signedvalue, $signedkey, $input) {
...
+	$rel_path = substr($absoltePath, $prefix_pos + strlen($root_directory));
```
joebordes, line 3911: prefix_pos could be `false` if strpos doesn't find it, not sure how this sum will act in that case

include/utils/utils.php
```diff
@@ -3904,4 +3904,12 @@ function validateClickHouseSecret($signedvalue, $signedkey, $input) {
...
+	return $rel_path;
```
joebordes, line 3912: we should return the value directly, there is no need to save the value in a variable just to return it

PR #1489, opened 2023-02-02
feat(Paint2Document) Ability to re-draw a document

asubaruwrxsti:


Smarty/templates/Components/Paint/index.tpl
```diff
@@ -64,14 +64,18 @@
 			<div class="slds-form-element__control">
 				<select id="folders" class="slds-input">
 					{foreach from=$FOLDERS item=name}
-					<option value="{$name[0]}"{if $FOLDERID==$name[0]} selected{/if}>{$name[1]}</option>
+						{if !($DOCFOLDERID)}
```
joebordes, line 67: please put this in one line, I think it is easier to read and maintain like this:

`<option value="{$name[0]}"{if $FOLDERID==$name[0] || $DOCFOLDERID==$name[0]} selected{/if}>{$name[1]}</option>`

PR #1498, opened 2023-02-09
Hide label related module related fields

maliknajjar:
# Create GV to hide the label of the related module in the related fields in the filter

Create the Application_Filter_Remove_RelatedModule_Label GV
When the GV is set to 1, when retrieving fields from a related module in the filter, the related module label is not shown.
For example, when the GV is set to 1, the labels of this filters should be:

Ente instead of Ente(Richiesta di approvazione) and Book instead of Book(Richiesta di approvazione)

## before
image

## after
image


include/ListView/ListViewGrid.php
```diff
@@ -429,7 +431,7 @@ public function findRelatedModule($fName) {
 	}
 
 	public function processResults($result, $field_types, $parentid = 0) {
-		global $adb, $current_user;
+		global $adb, $current_user, $current_module;
```
joebordes, line 434: is the $current_module variable necessary?

modules/GlobalVariable/language/en_us.gvdefs.php
```diff
@@ -1690,6 +1690,12 @@
 	'values' => '0 | 1',
 	'definition' => 'Edit the filter "All". The default is 1',
 ),
+'Application_Filter_Remove_RelatedModule_Label' => array(
+	'valuetype' => 'Boolean',
+	'category' => 'Application',
+	'values' => '0 | 1',
+	'definition' => 'Remove Related Module Label from column name',
```
joebordes, line 1697: change this to `Remove Related Module Label from table/list view column names`

PR #1505, opened 2023-02-17
Fixed crmEntity query for FTP

asubaruwrxsti:


joebordes:
Hi

Please do this PR again using pquery instead of convert2Sql and be careful with the formatting.

Thanks

PR #1509, opened 2023-02-20
feat(Expressions) support sha256 hash in workflow expression

maliknajjar:
support sha256 hash in workflow expression


joebordes:
Thanks!

PR #1510, opened 2023-02-20
feat(WorflowExpression) adding hashFile function to workflow Expression

maliknajjar:
adding hashFile function to workflow Expression to hash the content of a file path in WorfkowExpression environment


joebordes:
thanks!

PR #1523, opened 2023-02-24
fix(ListViewExport) lisView does not support exporting fields from other modules

maliknajjar:
* fixed the issue by using the lisview generated query instead creating it again
* supported features listed in the second image
* adding a notice message if the user used the search
* removed the code that is not used any more
image
image

### DIFF
--- a/data/CRMEntity.php
+++ b/data/CRMEntity.php
@@ -3683,11 +3683,11 @@ class CRMEntity {
 		$order_by = '';
 		$customView = new CustomView($cmodule);
 		$viewid = $customView->getViewId($cmodule);
-		$sortfieldbyfirst = !empty(cbCVManagement::getFieldValuesByCvId($viewid)) ? cbCVManagement::getFieldValuesByCvId($viewid)['sortfieldbyfirst'] : '';
+		$sortfieldbyfirst = !empty(cbCVManagement::getFieldValuesByCvId($viewid)) ? cbCVManagement::getFieldValuesByCvId($viewid) : '';
 		if (GlobalVariable::getVariable('Application_ListView_Default_Sorting', 0, $cmodule)) {
 			$order_by = GlobalVariable::getVariable('Application_ListView_Default_OrderField', $this->default_order_by, $cmodule);
-		} elseif (!GlobalVariable::getVariable('Application_ListView_Default_Sorting', 0, $cmodule) && !empty($sortfieldbyfirst)) {
-			$order_by = $sortfieldbyfirst;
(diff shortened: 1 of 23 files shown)


data/CRMEntity.php
```diff
@@ -3758,7 +3683,7 @@ public function getOrderBy() {
 		$order_by = '';
 		$customView = new CustomView($cmodule);
 		$viewid = $customView->getViewId($cmodule);
-		$sortfieldbyfirst = cbCVManagement::getFieldValuesByCvId($viewid)['sortfieldbyfirst'];
```
joebordes, line 3761: I fixed this warning a few days ago like this:  https://github.com/tsolucio/corebos/commit/2e0a1cbf43b71fed827c8bf90e517d25883a3552   which is correct/better?

include/utils/export.php
```diff
@@ -97,115 +97,58 @@ function export($type, $format = 'CSV') {
...
-			$query .= " and $focus->table_name.$focus->table_index in (" . generateQuestionMarks($idstring) . ')';
-			$params[] = $idstring;
-			// END
+	if ($export_data == 'currentpage') {
+		$query .= $_SESSION['limitQuery'];
+	} elseif ($export_data == 'selecteddata') {
+		$idsArray = explode(';', vtlib_purify($_REQUEST['idstring']));
+		$where = '';
+		foreach ($idsArray as $key => $value) {
+			if (!empty($value)) {
```
joebordes, line 112: why isnt this an `in ('. generateQuestionMarks($idstring) .')` ?  better as a list of ORs?

modules/com_vtiger_workflow/com_vtiger_workflow.php
```diff
@@ -25,62 +25,6 @@ class com_vtiger_workflow extends CRMEntity {
 	public function track_view($user_id, $current_module, $id = '') {
 	}
 
-	public function create_export_query($where) {
```
joebordes, line 28: does this work?

include/utils/export.php
```diff
@@ -105,6 +105,9 @@ function export($type, $format = 'CSV') {
 
 	if (method_exists($focus, 'create_export_query')) {
 		$query = $focus->create_export_query('');
+		global $log;
+		$log->fatal('PPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPP');
+		$log->fatal($query);
```
joebordes: :-)

maliknajjar: I removed them 👍

PR #1524, opened 2023-02-28
fix(FieldDependency) collapse a row only when all its fields are hidden

denaldhushi:


joebordes:
Thanks!!  This was important :-)

PR #1531, opened 2023-03-07
feat(slider) adding external link and download as zip to slider

maliknajjar:
adding external link and download as zip to slider


Smarty/templates/BrowserVariables.tpl
```diff
@@ -26,6 +26,7 @@ var gcbpageisdirty = false;
 var gcbpagewhyisdirty = '';
 var goldcorebos_browsertabID = '{$corebos_browsertabID}';
 var gVTviewType = '{$gVTviewType}';
+var gVTsiteUrl = '{$gVTsiteUrl}';
```
joebordes, line 29: this should be a separate commit: try to make small and semantically cohesive commits. this one, with the setting below, is an individual feature that can be used in the whole application, it is not semantically related to this new feature even if it does use it

Smarty/templates/sliderwidget.tpl
```diff
@@ -15,6 +15,7 @@
  *************************************************************************************************/
 -->*}
 <script src="include/components/Carousel/vanilla-js-carousel.min.js"></script>
+<script src="include/components/JSZip/JSZip.min.js"></script>
```
joebordes, line 18: I would prefer that we avoid this extra library. Can we get that zipped information from coreBOS?  I see below that you are fetch(ing) a URL which I suppose is to the coreBOS install to get the images, I would prefer that coreBOS sent it already compressed. Not only do we have all that functionality already in coreBOS but we will optimize the bandwidth and local browser memory.

include/components/slider.js
```diff
@@ -101,6 +101,41 @@ const Slider = {
...
+	const result = await ExecuteFunctions('shareLink', 'recordId=1084&operation=create');
+	const resultObj = JSON.parse(result);
+	const shareToken = resultObj['shareToken'];
+	const validUntil = resultObj['validUntil'];
+	const imageSliderShareLinkPopupContainer = document.getElementById('imageSliderShareLinkPopupContainer');
+	const imageSliderValidUntill = document.getElementById('imageSliderValidUntill');
+	const imageSliderShareLink = document.getElementById('imageSliderShareLink');
+	imageSliderShareLinkPopupContainer.style.display = 'flex'
+	const shareLink = `${gVTsiteUrl}/notifications.php?type=docshare&share_token=${shareToken}`;
+	imageSliderShareLink.innerHTML = `Your share link is: <a id="imageSliderShareLinkATag" href="${shareLink}">${shareLink}</a>`;
```
joebordes, line 114: have to translate these strings

include/js/general.js
```diff
@@ -7718,6 +7718,21 @@ function cbdzdragFinish() {
 	document.getElementById('corebosdropzonemsg').style.display = 'none';
 }
 
+async function downloadFilesAsZip(fileName, urls) {
+	var baseurl = `index.php?module=Utilities&action=UtilitiesAjax&file=ExecuteFunctions&functiontocall=compressAndDownloadFiles&zipFileName=test.zip&filesToBeCompressed=${JSON.stringify(urls)}`;
```
joebordes, line 7722: just to be sure you know that you can call this with the [ExecuteFunctions](https://github.com/tsolucio/corebos/blob/master/include/js/vtlib.js#L404) function.

modules/Documents/ShareDocLinkWidgetDetailViewBlock.php
```diff
@@ -134,10 +137,104 @@ public function sendFile($docid) {
...
+			$dataIMG = array();
+		}
+		$recordImages = cbws_getrecordimageinfo($docid, $current_user);
+		foreach ($recordImages['images'] as $key => $value) {
+			$dataIMG[] = $site_URL . '/' . $value['path'] . $value['id'] . '_' . $value['name'];
+		}
+	}
+
+	if (empty($dataIMG)) {
+		echo "there are no images to show";
```
joebordes, line 221: i18n

include/js/general.js
```diff
@@ -7718,6 +7718,21 @@ function cbdzdragFinish() {
...
+	fetch(baseurl)
```
joebordes, line 7723: just to be sure you know that you can call this with the [ExecuteFunctions](https://github.com/tsolucio/corebos/blob/master/include/js/vtlib.js#L404) function.

modules/Documents/ShareDocLinkWidgetDetailViewBlock.php
```diff
@@ -134,10 +137,104 @@ public function sendFile($docid) {
...
+	global $current_user, $adb, $site_URL, $log;
+	require_once 'modules/Utilities/cbSliderWidget.php';
+	require_once 'include/utils/utils.php';
+	require_once 'Smarty_setup.php';
+
+	$smq->clearAllExpiredShareLinksForSlider();
+	$share_token = vtlib_purify($_REQUEST['share_token']);
+	$tokenObject = json_decode(coreBOS_Settings::getSetting("slider_" . $share_token, 0), true);
+	if (empty($tokenObject)) {
+		echo "This token doesnt exist or it is expired";
```
joebordes, line 183: i18n

Smarty/templates/shareLinkImageSlider.tpl
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Image slider</title>
```
joebordes, line 7: i18n

include/js/general.js
maliknajjar: I tried it that way and converted the response into a blob but it the zip file wasn't opening. but using the fetch direcly it works
image

joebordes:
thanks!

PR #1538, opened 2023-03-20
Tui grid filter

asubaruwrxsti:


joebordes:
@denaldhushi that LGTM but I would like your review also as you are much more familiar with TUIGrid code.

PR #1545, opened 2023-03-24
fix(utilities)filter uitype10, save IDs, trigger activate when toggle…

ketlinapuci:
… from disabled to enabled


joebordes:
thanks!

PR #1547, opened 2023-03-27
Underscore not working in search

maliknajjar:
When trying to search the string "_text" it will consider the underscore as a wildcard that represents any value. in order to fix this we need to escape the underscore in the string.

### DIFF
--- a/include/QueryGenerator/QueryGenerator.php
+++ b/include/QueryGenerator/QueryGenerator.php
@@ -508,6 +508,7 @@ class QueryGenerator {
 	}
 
 	public function getQuery($distinct = false, $limit = '') {
+		global $adb;
 		if (empty($this->query)) {
 			$allFields = array_merge($this->whereFields, $this->fields);
 			foreach ($allFields as $fieldName) {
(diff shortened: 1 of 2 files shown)


maliknajjar:
@joebordes yes .. actually, I think maybe its better to even escape the $value variable itself. because we should not give the user the ability to use the underscore wildcard. and I don't think that query generator use it anywhere. wdyt?

joebordes:
> @joebordes yes .. actually, I think maybe its better to even escape the $value variable itself. because we should not give the user the ability to use the underscore wildcard. and I don't think that query generator use it anywhere. wdyt?

I don't agree because when they do an exact search we will want to have that underscore. I mean if someone is using the `equals` or `not equals`, I think we need to permit that underscore "as is", no?

maliknajjar:
yes I agree. because the "_" wildcard works only when used with the LIKE keyword. so how about escaping the underscore when the LIKE keyword is used?

joebordes:
yes, correct. that is the idea. I shared a link to the QG code section where starts/ends/contains are, there are probably some others there that need that escape

maliknajjar:
ok, I will add it to them too

joebordes:
LGTM. I'll wait to see the unit tests before merging. Ok?

maliknajjar:
yes .. I will create them

joebordes:
thanks!

PR #1553, opened 2023-03-31
Backward compatible new export system

maliknajjar:
the export feature does not work when we have a column from another module
this new export system will fix that. and in order to make it backward compatible it will be used only if you added a foreign module column. and it will use the old one if you did not add a foreign column.


joebordes:
merged in to NG

